PR #9, opened 2022-04-02
[SingleSource][Polybench] Add parentheses to make print_element behave as expected

jupiter-zzp:
[SingleSource][Polybench] Add parentheses to make print_element behave as expected


kbeyls:
Thank you for submitting this patch for review @jupiter-zzp .
The LLVM community does not use pull requests. Instead, it uses phabricator as documented on https://llvm.org/docs/Contributing.html#how-to-submit-a-patch and https://llvm.org/docs/Phabricator.html.
Please follow the guidelines there to submit your patch.

PR #20, opened 2023-09-08
[SingleSource/Vectorizer] Add unit tests for the vplan-native path.

iamlouk:
This patch adds unit tests for LLVM's VPlan-native path, as requested in https://reviews.llvm.org/D157484.

The new Vectorizer/VPlanNativePath subdirectory is only enabled if the compiler is clang. For all source files in that directory, the flags "-mllvm -enable-vplan-native-path" are added.

Four different scenarios are tested for outer-loop vectorization:
- Matrix multiplication, where the second of three loops is vectorized.
- A test where the vectorized loop has an auxiliary induction variable.
- A test for indirect and strided memory accesses.
- A nesting of three loops where the outer-most one is vectorized.

### DIFF
--- a/SingleSource/UnitTests/Vectorizer/VPlanNativePath/outer-loop-vect.cpp
+++ b/SingleSource/UnitTests/Vectorizer/VPlanNativePath/outer-loop-vect.cpp
@@ -1,0 +1,139 @@
+#include <cstddef>
+#include <cstdint>
+#include <iostream>
+#include <memory>
+
+#include "../common.h"
+
+// Tests for outer-loop vectorization in LLVM's VPlan-native path.
+
+#define DEFINE_SCALAR_AND_VECTOR_FN_FOR_OLV(Args, Loop)                       \
+  auto ScalarFn = [] Args {                                                   \
+    _Pragma("clang loop vectorize(disable) interleave_count(1)") Loop         \
+  };                                                                          \
+  auto VectorFn = [] Args {                                                   \
+    _Pragma("clang loop vectorize(enable)") Loop                              \
+  };
+
+#define DEFINE_SCALAR_AND_VECTOR_FN_FOR_NESTED_OLV(Args, Loop)                \
+  auto ScalarFn = [] Args {                                                   \
+    for (size_t i = 0; i < N; i++) {                                          \
+      _Pragma("clang loop vectorize(disable) interleave_count(1)") Loop       \
+    }                                                                         \
+  };                                                                          \
+  auto VectorFn = [] Args {                                                   \
+    for (size_t i = 0; i < N; i++) {                                          \
+      _Pragma("clang loop vectorize(enable)") Loop                            \
+    }                                                                         \
+  };
+
+int main() {
+  rng = std::mt19937(15);
+
+  {
+    // A matrix-multiplication where the second loop of the triple loop nest
+    // is vectorized (the macro adds the outer-most loop).
+    DEFINE_SCALAR_AND_VECTOR_FN_FOR_NESTED_OLV(
+      (size_t N, size_t M, size_t L,
+       int32_t *__restrict__ A, const int32_t *B, const int32_t *C),
+      for (size_t j = 0; j < L; j++) {
+        int32_t a = 0;
+        for (size_t k = 0; k < M; k++)
+          a += B[i*M+k] * C[k*L+j];
+        A[i*L+j] = a;
+      });
+
+    std::cout << "Checking matrix-multiplication\n";
+
+    size_t N = 100, M = 100, L = 100;
+    std::unique_ptr<int32_t[]> A_Reference(new int32_t[N*L]);
+    std::unique_ptr<int32_t[]> A_ToCheck(new int32_t[N*L]);
+    std::unique_ptr<int32_t[]> B(new int32_t[N*M]);
+    std::unique_ptr<int32_t[]> C(new int32_t[M*L]);
+    init_data(B, N*M);
+    init_data(C, M*L);
+
+    ScalarFn(N, M, L, &A_Reference[0], &B[0], &C[0]);
+    VectorFn(N, M, L, &A_ToCheck[0], &B[0], &C[0]);
+    check(A_Reference, A_ToCheck, N*L);
+  }
+
+  {
+    // A test where the vectorized loop itself has an auxiliary IV.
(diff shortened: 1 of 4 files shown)


iamlouk:
I do not seam to be able to add reviewers, so sorry for tagging you here in a comment: @fhahn @alexey-bataev 

Please also note that the matrix multiplication test will cause current LLVM to crash without https://reviews.llvm.org/D150700 which is not yet merged/accepted.

fhahn:
Sorry for the delay with reviewing this!

Trying to build this with the latest version of `main` triggers an assert:

```
Assertion failed: ((From->getParent() == To->getParent()) && "Can't connect two block with different parents"), function connectBlocks, file VPlan.h, line 2840.
```

I'll get a reproducer.

iamlouk:
No worries, thank you for having a look. I cannot confirm if its exactly that assert that breaks because I do not have my computer near by, but I know that the bug fixed in this patch is triggered by one of the tests: https://reviews.llvm.org/D150700 ("[LV] Stability fix for outerloop vectorization") by @nikolaypanchenko (@npanchen ? Sorry if tagging a random stranger, GitHub suggests this username).

fhahn:
> No worries, thank you for having a look. I cannot confirm if its exactly that assert that breaks because I do not have my computer near by, but I know that the bug fixed in this patch is triggered by one of the tests: https://reviews.llvm.org/D150700 ("[LV] Stability fix for outerloop vectorization") by @nikolaypanchenko (@npanchen ? Sorry if tagging a random stranger, GitHub suggests this username).

Indeed, looks like the patch should help, but would need a rebase.

fhahn:
@iamlouk could you check if the new tests now build and run without issues after https://github.com/llvm/llvm-project/pull/68118 landed?

iamlouk:
Hello, yes, the tests compile and pass now!

I tested with llvm-project commit 88f0e4c75c1ac498f2223fc640c4ff6c572c5ed1, the llvm-test-suite was built using `cmake -DCMAKE_C_COMPILER=<...>/bin/clang -DCMAKE_CXX_COMPILER=<...>/bin/clang++ -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_FLAGS="-mcpu=native" -DCMAKE_CXX_FLAGS="-mcpu=native" ..` on a AArch64 host with SVE.

SingleSource/UnitTests/Vectorizer/VPlanNativePath/outer-loop-vect.cpp
```diff
@@ -0,0 +1,139 @@
...
+int main() {
+  rng = std::mt19937(15);
+
+  {
+    // A matrix-multiplication where the second loop of the triple loop nest
+    // is vectorized (the macro adds the outer-most loop).
+    DEFINE_SCALAR_AND_VECTOR_FN_FOR_NESTED_OLV(
+      (size_t N, size_t M, size_t L,
+       int32_t *__restrict__ A, const int32_t *B, const int32_t *C),
+      for (size_t j = 0; j < L; j++) {
```
fhahn, line 39: Would be good to consistently use upper case for variable names, as per LLVM coding style.

fhahn (approved):
LGTM, thanks!

An additional suggestion about keeping the variable names consistent inline.

SingleSource/UnitTests/Vectorizer/VPlanNativePath/outer-loop-vect.cpp
iamlouk: Sorry for this, I fixed that!

fhahn (approved):
Thanks for the latest update, still LGTM!

iamlouk:
Sorry to ask you for this, but could you merge this for me? I don't have the permissions. Thank you very much in advance.

fhahn:
@SixWeining interesting, could you share the IR causing the crash? (`-mllvm -print-on-crash -mllvm -print-module-scope`)

SixWeining:
> @SixWeining interesting, could you share the IR causing the crash? (`-mllvm -print-on-crash -mllvm -print-module-scope`)

Uploaded: [crash.ll.txt](https://github.com/llvm/llvm-test-suite/files/14031351/crash.ll.txt)  Thanks.

SixWeining:
Note that llvm is built with `LLVM_ENABLE_EXPENSIVE_CHECKS=ON`. The `OFF` build doesn't have such issue.

fhahn:
@SixWeining not sure how expensive checks impact the codegen, but the issue should be fixed by a04f6152914ea21f3068aaba9d8fc21d2e703d3e

SixWeining:
> @SixWeining not sure how expensive checks impact the codegen, but the issue should be fixed by a04f6152914ea21f3068aaba9d8fc21d2e703d3e

Unfortunately, still crash with the `assert` in `llvm/lib/Transforms/Vectorize/LoopVectorize.cpp` and this assert is only enabled on a `LLVM_ENABLE_EXPENSIVE_CHECKS=ON` build.
```
 3141 #ifdef EXPENSIVE_CHECKS
 3142   assert(DT->verify(DominatorTree::VerificationLevel::Fast));
 3143 #endif
 3144 
 3145   return LoopVectorPreHeader;
 3146 }
```

iamlouk:
Hello, I think this crash is just because of [this condition in `VPlan::execute()`](https://github.com/llvm/llvm-project/blob/df5e431e8dcb555320be7f7ae4d499a11e0b17db/llvm/lib/Transforms/Vectorize/VPlan.cpp#L843): `// We do not attempt to preserve DT for outer loop vectorization currently.`. I don't think this is an actual problem, though, because the dominator tree is not a preserved analysis in case the VPlan-native path is enabled ([see here](https://github.com/llvm/llvm-project/blob/df5e431e8dcb555320be7f7ae4d499a11e0b17db/llvm/lib/Transforms/Vectorize/LoopVectorize.cpp#L10326)). IHMO, it would be fine just adding a `if (!EnableVPlanNativePath)` in front of the assert highlighted by @SixWeining .

@fhahn , the changes you made in [a04f6152](https://github.com/llvm/llvm-project/commit/a04f6152914ea21f3068aaba9d8fc21d2e703d3e) are what could be needed to be able to remove the "-fno-vectorize" I added [here](https://github.com/iamlouk/llvm-test-suite/blob/1382bbc0caadce460cd1740dc12c8376d055bb32/SingleSource/UnitTests/Vectorizer/VPlanNativePath/CMakeLists.txt#L4), but have not confirmed that yet.

PR #21, opened 2023-09-08
[SingleSource] Pass -fenable-matrix at link time in case ThinLTO is used

w2yehia:
When running the test-suite with -flto=thin, `matrix-types-spec.test` fails because matrix intrinsics are not lowered at compile time with -flto=thin.
With, [D153583](https://reviews.llvm.org/D153583), the compiler lowers the intrinsic at link time when -fenable-matrix and -flto=thin is specified.

### DIFF
--- a/SingleSource/UnitTests/CMakeLists.txt
+++ b/SingleSource/UnitTests/CMakeLists.txt
@@ -65,3 +65,7 @@ else()
 endif()
 
 llvm_singlesource()
+
+if (COMPILER_HAS_MATRIX_FLAG)
+  set_property(TARGET matrix-types-spec PROPERTY LINK_OPTIONS -fenable-matrix)
+endif()


w2yehia:
@fhahn @AaronBallman ping. Thanks

SingleSource/UnitTests/CMakeLists.txt
```diff
@@ -65,3 +65,7 @@ else()
 endif()
 
 llvm_singlesource()
+
+if (COMPILER_HAS_MATRIX_FLAG)
+  set_property(TARGET matrix-types-spec PROPERTY LINK_OPTIONS -fenable-matrix)
```
fhahn, line 70: Can this be moved to the `if` above?

fhahn (commented):
Thanks for the fix!

SingleSource/UnitTests/CMakeLists.txt
fhahn: Thanks for checking!

fhahn (approved):
LGTM, thanks!

PR #25, opened 2023-09-19
[test-suite][Fortran] Fix failing tests from gfortran test suite

tarunprabhu:
Disable two tests that fail when optimizations are enabled, but pass otherwise. Comments have been added in the two DisabledFiles.cmake files describing why the tests were disabled.

These failures were reproduced by @kiranchandramohan  and @cabreraam on Slack.

### DIFF
--- a/Fortran/gfortran/regression/DisabledFiles.cmake
+++ b/Fortran/gfortran/regression/DisabledFiles.cmake
@@ -2322,6 +2322,12 @@ file(GLOB FAILING_FILES CONFIGURE_DEPENDS
   write_check.f90
   zero_sized_1.f90
 
+  # ---------------------------------------------------------------------------
+  #
+  # This test fails with optimizations enabled, but succeeds when compiled
+  # without optimizations.
+  inline_transpose_1.f90
+
   # These tests fail at runtime on AArch64 (but pass on x86). Disable them
   # anyway so the test-suite passes by default on AArch64.
   entry_23.f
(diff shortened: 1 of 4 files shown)


kiranchandramohan (approved):
LG. Thanks, will you create a ticket for `inline_transpose_1.f90`?

tarunprabhu:
> LG. Thanks, will you create a ticket for `inline_transpose_1.f90`?

I suppose that would depend on how soon HLFIR will be enabled by default. Since @tblah doesn't see the error with HLFIR, it may be moot once that is enabled. If you think I should anyway, I'm happy to. 

> After this change, `getarg_1.f90` can be removed from `Fortran/gfortran/torture/execute/DisabledFilesHLFIR.cmake`

Do you mean `inline_transpose_1.f90` here? `getarg_1.f90` will overflow the string if the executable name is longer than 10 characters, so it should be disabled as long as the build system generates executable names longer than that.

> Would you like me to make these HLFIR changes in a separate patch or will you include them in this one?

I'll add it to this patch. Thanks.

tarunprabhu:
> > After this change, `getarg_1.f90` can be removed from `Fortran/gfortran/torture/execute/DisabledFilesHLFIR.cmake`
> 
> Do you mean `inline_transpose_1.f90` here? `getarg_1.f90` will overflow the string if the executable name is longer than 10 characters, so it should be disabled as long as the build system generates executable names longer than that.

Ah, never mind. I see that it was disabled for HLFIR already. Since it is now disabled in the main deny list, there is no reason to leave it in the HLFIR deny list.

tblah:
Thanks for the update. Did you intend to add `update-disabled-tests.py`? I love the idea of the script, but it seems to be incomplete.

cabreraam:
> Thanks for the update. Did you intend to add `update-disabled-tests.py`? I love the idea of the script, but it seems to be incomplete.

+1

Just by reading the title, it seems very useful.

cabreraam (approved):

tarunprabhu:
Whoops. No I didn't. Committed too hastily, I'm afraid. 

Yes, it is something I have been working on and it is incomplete.

PR #26, opened 2023-09-27
[MicroBenchmark,LoopInterleaving] Check performance impact of Loop Interleaving Count with varying loop iterations

nilanjana87:
This microbenchmark attempts to find the right loop trip count threshold for deciding whether to interleave a loop or not for different types of loops, such as loops with or without reduction inside it, loops with or without vectorization inside it. For context, the loop vectorizer uses a threshold of [TinyTripCountInterleaveThreshold](https://github.com/llvm/llvm-project/blob/56f7c7e52fe34fefda217b1194f681333159a809/llvm/lib/Transforms/Vectorize/LoopVectorize.cpp#L5672) that is currently set to 128.


MicroBenchmarks/LoopVectorization/LoopInterleaving.cpp
```diff
@@ -0,0 +1,313 @@
+// This program tests performance impact of Interleaving Count with varying loop
+// iteration count for different types of loop, such as loops with or
+// without reduction inside it, loops with different vectorization widths.
```
fhahn, line 3: nit: `reductions` inside.

MicroBenchmarks/LoopVectorization/LoopInterleaving.cpp
```diff
@@ -0,0 +1,313 @@
...
+  for (auto _ : state) {
+    benchmark::DoNotOptimize(A);
+    benchmark::DoNotOptimize(B);
+    benchmark::DoNotOptimize(C);
+    benchmark::ClobberMemory();
+    g_sum += Fn(Iterations);
+  }
+}
+
+#define ADD_BENCHMARK(Itr)                                                     \
```
fhahn, line 142: Could you document what this is adding?

MicroBenchmarks/LoopVectorization/LoopInterleaving.cpp
```diff
@@ -0,0 +1,313 @@
...
+  for (int J = 0; J < Iterations; J++) {
+    sum += A[J];
+  }
+  return sum;
+}
+
+static void __attribute__((always_inline))
+benchForLoopInterleaveThreshold(benchmark::State &state, void (*Fn)(int),
+                                int Iterations) {
+  std::uniform_int_distribution<int> distrib(std::numeric_limits<int>::min(),
```
fhahn, line 115: unused?

MicroBenchmarks/LoopVectorization/LoopInterleaving.cpp
```diff
@@ -0,0 +1,313 @@
...
+loopWithReductionWithVW1IC2(int Iterations) {
+  unsigned sum = 0;
+#pragma clang loop vectorize_width(1) interleave_count(2)
+  for (int J = 0; J < Iterations; J++) {
+    sum += A[J];
+  }
+  return sum;
+}
+
+static int __attribute__((noinline))
```
fhahn, line 102: It would probably also be good to have a variant that lets the vectorizer automatically pick UF & VF

MicroBenchmarks/LoopVectorization/LoopInterleaving.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+static void __attribute__((noinline)) loopWithVW4IC1(int Iterations) {
+#pragma clang loop vectorize_width(4) interleave(disable)
+  for (int J = 0; J < Iterations; J++) {
+    A[J] = B[J] + C[J];
+  }
+}
+
+static void __attribute__((noinline)) loopWithVW4IC2(int Iterations) {
+#pragma clang loop vectorize_width(4) interleave_count(2)
```
fhahn, line 39: Not sure if that's possible, but could the `vectorize_with` and `internalave_count` be template arguments?

fhahn:
IIUC this adds benchmarks for the best case for interleaving. Would it be possible to also add variants where interleaving is less beneficial?

MicroBenchmarks/LoopVectorization/LoopInterleaving.cpp
nilanjana87: Since the vectorization hints are added via the `#pragma` directive, the `#` inside the macro function definition is causing problem & I couldn't figure out a way to escape the `#` even with macro expansion.

MicroBenchmarks/LoopVectorization/LoopInterleaving.cpp
```diff
@@ -0,0 +1,313 @@
+// This program tests performance impact of Interleaving Count with varying loop
+// iteration count for different types of loop, such as loops with or
```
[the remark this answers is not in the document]
nilanjana87: done.

MicroBenchmarks/LoopVectorization/LoopInterleaving.cpp
nilanjana87: Merged the functionality in one function & deleted this copy.

MicroBenchmarks/LoopVectorization/LoopInterleaving.cpp
nilanjana87: Added new tests using functions with `WithoutVecHint` included in the name, that will let the vectorizer pick the UF & VF.

nilanjana87:
> IIUC this adds benchmarks for the best case for interleaving. Would it be possible to also add variants where interleaving is less beneficial?

This file adds benchmarks for testing cases that may or may not be beneficial for interleaving. For example, cases with low trip counts are better off without interleaving, whereas if the interleaved loop runs at least twice it starts showing performance benefit. If we need to add more cases where it is less beneficial, can you suggest some?

fhahn:
> > IIUC this adds benchmarks for the best case for interleaving. Would it be possible to also add variants where interleaving is less beneficial?
> 
> This file adds benchmarks for testing cases that may or may not be beneficial for interleaving. For example, cases with low trip counts are better off without interleaving, whereas if the interleaved loop runs at least twice it starts showing performance benefit. If we need to add more cases where it is less beneficial, can you suggest some?

Some additional cases could be loops with a reduction, but also a number of additional independent memory and compute operation chains (i.e. larger loop body with multiple independent operation chains) and one with a larger loop body without reduction (in that case, LLVM likely will decide not to interleave).

They might provide additional insight into current issues of the cost modeling for interleaving.

MicroBenchmarks/LoopVectorization/LoopInterleaving.cpp
nilanjana87: Figured a way to do this using the `_Pragma` operator.

nilanjana87:
> > > IIUC this adds benchmarks for the best case for interleaving. Would it be possible to also add variants where interleaving is less beneficial?
> > 
> > 
> > This file adds benchmarks for testing cases that may or may not be beneficial for interleaving. For example, cases with low trip counts are better off without interleaving, whereas if the interleaved loop runs at least twice it starts showing performance benefit. If we need to add more cases where it is less beneficial, can you suggest some?
> 
> Some additional cases could be loops with a reduction, but also a number of additional independent memory and compute operation chains (i.e. larger loop body with multiple independent operation chains) and one with a larger loop body without reduction (in that case, LLVM likely will decide not to interleave).
> 
> They might provide additional insight into current issues of the cost modeling for interleaving.

Replicated the existing tests for loops with bigger bodies with additional independent memory operations. These can be found in the latest patch, if you do case-insensitive search of function names for "BigLoop".

fhahn (commented):
Thanks for adding the extra loops, a few more comments inline

MicroBenchmarks/LoopVectorization/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ llvm_test_executable(LoopVectorizationBenchmarks
   MathFunctions.cpp
   RuntimeChecks.cpp
   VectorOperations.cpp
+  LoopInterleaving.cpp
```
fhahn, line 17: It would probably be better to have this as a separate benchmark binary, as here are a lot of benchmarks added in this file and running it takes a while

MicroBenchmarks/LoopVectorization/LoopInterleaving.cpp
```diff
@@ -0,0 +1,399 @@
...
+    g_sum += Fn(Iterations);
+  }
+}
+
+#define STRINGIFY(a) #a
+
+// Loops without Reduction with different vectorization configurations
+
+static int __attribute__((noinline)) loopNoReductionAutoVec(int Iterations) {
+  for (int J = 0; J < Iterations; J++) {
```
fhahn, line 60: Could you double check that this loop doesn't get unrolled in case LLVM can propagate the constant iterations counts through to the function (may happen even with `no inline`)

nilanjana87: I checked that it doesn't unroll the loop because of propagating constant iteration counts & specifically disabled loop unrolling for the auto-vectorization cases.

fhahn (approved):
LGTM, thanks!

antmox:
Hello. I think this patch broke several bots:
- clang-aarch64-lld-2stage https://lab.llvm.org/buildbot/#/builders/185/builds/5395
- clang-armv8-lld-2stage https://lab.llvm.org/buildbot/#/builders/178/builds/6117
- clang-aarch64-full-2stage https://lab.llvm.org/buildbot/#/builders/179/builds/8485
- clang-aarch64-sve-vla-2stage https://lab.llvm.org/buildbot/#/builders/198/builds/6328
- clang-aarch64-sve-vls-2stage https://lab.llvm.org/buildbot/#/builders/176/builds/7027
- clang-armv7-lnt https://lab.llvm.org/buildbot/#/builders/174/builds/27416
- clang-armv7-global-isel https://lab.llvm.org/buildbot/#/builders/186/builds/13219
- clang-aarch64-global-isel https://lab.llvm.org/buildbot/#/builders/183/builds/17403
- clang-aarch64-sve-vla https://lab.llvm.org/buildbot/#/builders/197/builds/10921

Could you please look at this ?

antmox:
Reverted it. Hope that it's OK and that it was indeed the culprit commit.

nilanjana87:
> Reverted it. Hope that it's OK and that it was indeed the culprit commit. (edit: yes it was)

Thanks for figuring this out & reverting it. Seems like adding too many tests ended up hitting the timeout. I'll try to reduce the test points to keep it within the time limit & re-land it. 

The tests passed for me locally though, even with the `ninja check` command that timed out in the build-bot. So, what is a good test to find this issue locally, other than reducing the test time below 1200 seconds as suggested by the failure log?

On a separate matter, I was checking the bot links you posted above, but I don't see myself or this patch in the `Responsible Users` or `Changes` tab, for example in https://lab.llvm.org/buildbot/#/builders/185/builds/5395. Is this because my patch has been removed from this list after it was reverted?

antmox:
Yes the test does completes successfully here as well. I don't know the rules for adding new tests, but this one seems really long.
2500sec here on a huge aarch64 machine, while the testsuite can run on smaller machines, sometimes heavily loaded.
I don't think there is a 1200sec limit, but jobs are killed after 1200sec without any output.

For the other thing, the responsible users and changes tabs should be read with care.
Changes from the llvm-test-suite and lnt repos are not listed here. Some bots may not even list all changes from the llvm-project here.
I think this is defined in the llvm-zorg builders classes, depends-on-projects attribute.

antmox:
Also, I wonder why only the arm/aarch64 bots failed with this test. Maybe there's something to analyze here.
On what type of machine did you run your test, and what was the execution time?

nilanjana87:
> 2500sec here on a huge aarch64 machine, while the testsuite can run on smaller machines, sometimes heavily loaded.
> I don't think there is a 1200sec limit, but jobs are killed after 1200sec without any output.

It runs for little more than 40min in my machine, but locally it doesn't hit the timeout. I added a new PR https://github.com/llvm/llvm-test-suite/pull/56 which runs a reduced test set by default and runs the whole set when compiled with a flag.

> Also, I wonder why only the arm/aarch64 bots failed with this test.
There's nothing Arm/AArch64 specific in the code, so I don't have a clue why this is so.

PR #29, opened 2023-10-03
[Polybench] Update to v4.2.1-beta

jerryzj:
- Old PolyBench/C 3.2 kernels had numerous implementation errors making their outputs to not match what is expected from the mathematical specification of the operation. Many of them did not influence the program behavior. (e.g., the number and type of operations, data
dependences, and overall control-flow was similar to the corrected implementation)

- For more details, please see the README

- Tested on macOS (clang-1500.0.40.1), RHEL (clang 14.0.6 and gcc 13.2.0) and RISCV usermode QEMU rv64gcv (clang upstream-116a1aeff)

- Source: https://sourceforge.net/projects/polybench/files/


fhahn:
Upgrading to the latest version makes sense to me in general. Could you link to the source in the PR? I guess there's no non-beta version of `v4.2.1`?

jerryzj:
@fhahn I've just updated the link for Polybench, but, unfortunately, there is no non-beta version available.

rengolin (approved):
This makes sense to me, too. Thanks for the update.

LGTM, but please wait for @fhahn's approval too, to make sure we're not missing anything.

fhahn:
> @fhahn I've just updated the link for Polybench, but, unfortunately, there is no non-beta version available.

Thanks! I am planning to give this a try with top-of-tree Clang tomorrow.

jerryzj:
@fhahn I have just finished testing upstream LLVM (116a1aeff) using RISCV target (rv64gcv/lp64d) under `O3` optimization, and I encountered 2 test failures.

```
test-suite.SingleSource/Benchmarks/Polybench/medley/deriche deriche.test
test-suite.SingleSource/Benchmarks/Polybench/medley/floyd-warshall floyd-warshall.test
```

I don't know if they also failed in your local testing?

fhahn:
> @fhahn I have just finished testing upstream LLVM (116a1aeff) using RISCV target (rv64gcv/lp64d) under `O3` optimization, and I encountered 2 test failures.
> 
> ```
> test-suite.SingleSource/Benchmarks/Polybench/medley/deriche deriche.test
> test-suite.SingleSource/Benchmarks/Polybench/medley/floyd-warshall floyd-warshall.test
> ```
> 
> I don't know if they also failed in your local testing?

All tests are passing on current `main` on ARM64 macOS. @jerryzj do the RISCV tests fail when building with `-O0`? If that's the case, then those new tests may just highlight genuine mis-compiles with RISCV.

jerryzj:
@fhahn yea, these test also failed with `O0`... 😞

SingleSource/Benchmarks/Polybench/README
```diff
@@ -1,9 +1,145 @@
...
+* New in 4.2:
+-------------
+ - Fixed a bug in syr2k.
+ - Changed the data initialization function of several benchmarks.
+ - Minor updates in the documentation and PolyBench API.
+
+-------------
+* New in 4.1:
+-------------
+ - Added LICENSE.txt
```
fhahn, line 84: It looks like `LICENSE.txt` has not be included? There's `SingleSource/Benchmarks/Polybench//LICENSE`, but that still mentions 3.2 and needs updating I think

fhahn (commented):
> @fhahn yea, these test also failed with `O0`... 😞

In a way, it already shows the value of the update! I don't think this needs to be a blocker; once merged, it would be great if you could file issues for the mis-compiles. Having this merged will also make it much easier for people to reproduce the failure.

I left one more inline comment regarding the license files.

SingleSource/Benchmarks/Polybench/README
jerryzj: Oops, I missed it. I just updated.

jerryzj:
@fhahn It seems that the tests failed because they hit the maximum output limit. I will try using a smaller problem size for them since the new version defaults to a rather large problem size.

jerryzj:
By default, lit has a constraint for output file size = 100MB

```
TIMEITFLAGS="$TIMEITFLAGS --limit-file-size 104857600"
```

While for deriche, the output file exceeds 100MB...

```
du -sh deriche.out
135M	deriche.out
```

jerryzj:
@fhahn @rengolin The tests are PASS on RISCV now~

jerryzj:
@fhahn If there are no further issues, I feel that we can merge this?

fhahn (approved):
LGTM, thanks!

PR #31, opened 2023-10-09
[Fortran][gfortran] respect CMAKE_Fortran_COMPILER_LAUNCHER

yuanfang-chen:
My use case is to limit long-running compilation (stuck, infinite loop). It should be useful for tools like ccache too.


tarunprabhu (approved):
LGTM. Thanks. This is a good idea.

PR #35, opened 2023-10-16
[External][SOLLVE V&V] add Fortran test cases

yuanfang-chen:
(it's hard to properly create stacked PR. I meant to commit #34 first, then this one)

### DIFF
--- a/External/sollve_vv/CMakeLists.txt
+++ b/External/sollve_vv/CMakeLists.txt
@@ -27,6 +27,261 @@ set(ALL_LIST_OPTIONS
   REGRESSION_REDLIST
 )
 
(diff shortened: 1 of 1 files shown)


ajarmusch:
this formating - is this from a formatter?

yuanfang-chen:
Only the last commit matters. The first two commits are from https://github.com/llvm/llvm-test-suite/pull/34

ajarmusch:
Have you run lit with these fortran tests? Do they passing?

yuanfang-chen:
I expect many of them to fail. So these are gated by `TEST_SUITE_FORCE_ALL`.

PR #38, opened 2023-10-18
[External]Add OpenACC Validation and Verification testsuite

wangxin0321:
OpenACC is a high-level directive-based parallel programming model that can manage the sophistication of heterogeneity in architectures and abstract it from the users. The portability of the model across CPUs and accelerators has gained the model a wide variety of users. This means it is also crucial to analyze the reliability of the compilers' implementations.

To address this challenge, the OpenACC Validation and Verification team introduces a validation testsuite ，supported by a streamlined infrastructure, to verify the OpenACC implementations across various compilers and system architectures.


erichkeane:
I have no idea how to review this unfortunately!  However I believe @sunitachandra showed interest on the Clang RFC, so would have a better idea if this is being done correctly.

sunitachandra:
Thanks @erichkeane! 
Adding my group's developer, Aaron Jarmusch, to help with the review @ajarmusch

ajarmusch:
@wangxin0321 Thanks for taking the time to add the OpenACC V&V to Externals

The header files are located in `Tests/`, so maybe that folder should be included instead of the repo root?

wangxin0321:
@ajarmusch 
Thank you for your suggestions. I have made the following modifications in response to your suggestions:
1 "set(_includedir "${TEST_SUITE_OpenACCVV_ROOT}/") and target_include_directories(${_name} PRIVATE) "${_includedir}")"(the addition of this header is not required),
At the same time, the following simple modifications were made
1 Change the file name OpenACCV_V.
2 Change the variable names of TEST_SUITE_OFFLOADING_FLAGS and TEST_SUITE_OFFLOADING_LDFLAGS.
3 Remove the "-master"

ajarmusch:
@wangxin0321 the changes look good

anyone have thoughts? otherwise, LGTM to merge

erichkeane:
So how are these configured to run on Clang?  Are they using the `-fopenacc` flag?  Do they depend on the `_OPENACC` macro in some way?

If so on the latter, can we have a build option to make it configurable with the `-fexperimental-openacc-macro-override` flag as well?

ajarmusch:
@erichkeane the flags are defined in the cmake command during the test suite build. The README gives an example using `-fopenacc`

External/OpenACCV_V/CMakeLists.txt
```diff
@@ -0,0 +1,70 @@
...
+    set(_name "acctargetvv-${_basename}.${_ext}")
+
+    llvm_test_run()
+
+    llvm_test_executable(${_name} "${TEST_SUITE_OpenACCVV_ROOT}/Tests/${_file}")
+     
+    target_link_libraries(${_name} PUBLIC OpenACC::OpenACC_${_lang} m)
+
+    # Add -fopenacc to linker command line; for some reason this is not done by target_link_libraries.
+    target_link_options(${_name} PRIVATE ${OpenACC_${LANG}_FLAGS})
```
erichkeane, line 41: So this comment isn't accurate, right?  It is adding the openacc-language flags, not the -fopenacc?  Or where are the `OpenACC_C_flags` and `OpenACC_CXX_flags` set?

erichkeane:
> @erichkeane the flags are defined in the cmake command during the test suite build. The README gives an example using `-fopenacc`

Right, but I'm doubting the usefulness of adding this suite WITHOUT that flag, since that flag is required to enable OpenACC anyway, right?  And I was hoping for a more 'builtin' way of using `-fexperimental-openacc-macro-override` (if the tests use `_OpenACC` in any way) in the Clang compiler, since we'd like to enable to this to validate our effort.

External/OpenACCV_V/CMakeLists.txt
ajarmusch: my fault - you were right before

ajarmusch:
good point, I looked at the sollve_vv and the `OPENMP_C_FLAG` flag is set in the OpenMP package. The sollve_vv uses `find_package(OpenMP)` which is in the root [CMakeList.txt](https://github.com/llvm/llvm-test-suite/blob/main/CMakeLists.txt). So as long as the idea is to replicate it for OpenACC that should be hardcoded. Would that work?

Also, you are hoping for a builtin option for `-fexperimental-openacc-macro-override`?  If so @wangxin0321 can you add the option for the flag?

erichkeane:
> good point, I looked at the sollve_vv and the `OPENMP_C_FLAG` flag is set in the OpenMP package. The sollve_vv uses `find_package(OpenMP)` which is in the root [CMakeList.txt](https://github.com/llvm/llvm-test-suite/blob/main/CMakeLists.txt). So as long as the idea is to replicate it for OpenACC that should be hardcoded. Would that work?
> 
> Also, you are hoping for a builtin option for `-fexperimental-openacc-macro-override`? If so @wangxin0321 can you add the option for the flag?

I would love something like that, exposed at a place that is easier to use/pipe in, I suspect this is something I'll use often.

External/OpenACCV_V/CMakeLists.txt
wangxin0321: "OpenACC_C_flags "and "OpenACC_CXX_flags" are set in cmake. Check out the official FindOpenACC documentation at the following website: https://cmake.org/cmake/help/latest/module/FindOpenACC.html

But my error, currently find_package(OpenACC), only NVHPC, PGI, GNU and Cray compilers are supported, We cannot use this method to see if the current llvm supports OpenACC. "OpenACC_C_flags "and "OpenACC_CXX_flags" are currently invalid

ajarmusch: @wangxin0321 Is there a way around? Or we have to make an issue with cmake?

ajarmusch:
@wangxin0321 Hi, It's been a while since we heard from you. Any progress so far?

External/OpenACCV_V/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+    get_filename_component(_ext "${_file}" EXT)
+    get_filename_component(_basename "${_file}" NAME_WE)
+    get_filename_component(_directory "${_file}" DIRECTORY)
+    string(REPLACE "." "" _ext "${_ext}")
+    set(_name "acctargetvv-${_basename}.${_ext}")
+
+    llvm_test_run()
+
+    llvm_test_executable(${_name} "${TEST_SUITE_OpenACCVV_ROOT}/Tests/${_file}")
+
```
ajarmusch, line 33:
```suggestion
     # Add -fopenacc to linker command line; for some reason this is not done by target_link_libraries.
    target_link_options(${_name} PRIVATE ${OpenACC_${LANG}_FLAGS})
```

ajarmusch:
Hi @wangxin0321, to move this along I am going to help. Since I am unable to make edits to your fork repository. I will create a separate PR and mention this PR and you as well for credit. Reach out as soon as possible

PR #40, opened 2023-10-20
[External][SOLLVE V&V] add Fortran test cases

yuanfang-chen:
Eventually, it's better to duplicate the list for each vendor. For now, it's an opt-in by `-DTEST_SUITE_FORCE_ALL=ON`.


kiranchandramohan:
Thanks @yuanfang-chen for this patch. Is the V&V test already inside the test repository? And is this patch only enabling them? Are these tests passing or failing now? Should that be checked and failing tests be disabled?

PR #42, opened 2023-10-31
[CLAMR] remove unused codes

jerryzj:
- The variable "progCL" is a l-value and never get used, but it will cause SEGV when argv (binary file path) is longer than its pre-defined length.


rengolin:
I can't remember why we created that but I think it's left-over from the time we had to handle with non-reproducible builds. If this still passes the test-suite on all platforms and isn't used, then it's fine to remove.

rengolin (approved):

jerryzj:
Yes, we can verify that these codes will be removed by the compiler when we enable optimizations. We encountered a SEGV only in `O0`.

PR #49, opened 2023-11-14
Port gcc-dg tests from GCC

mohammed-nurulhoque:
This was copied from a snapshot of gcc at 0c37fef39fa. The allowlist was tested using glibc and qemu on riscv.
Would appreciate any help regarding copyrights, what notices should be included


jerryzj:
Cool, I am going to test it in our downstream.

jerryzj (changes requested):
Build failed on my end, please consider add the following

```
git diff SingleSource/Regression/DgOptions.txt
diff --git a/SingleSource/Regression/DgOptions.txt b/SingleSource/Regression/DgOptions.txt
index 935899dfc..f951ce935 100644
--- a/SingleSource/Regression/DgOptions.txt
+++ b/SingleSource/Regression/DgOptions.txt
@@ -8,9 +8,13 @@ set(CLANG_ERRORING_CFLAGS
   "-fopt-"
   "-fno-ipa-"
   "-fno-ira-"
+  "-fno-ira-share-spill-slots"
   "-ftree-loop-"
+  "-ftree-loop-distribution"
   "-fno-tree-b"
+  "-fno-tree-bit-ccp"
   "-fno-tree-c"
+  "-fno-tree-coalesce-vars"
   "-fno-tree-d"
   "-fno-tree-f"
   "-fno-tree-t"
```

jerryzj:
In the `main` branch, `lit` has 2437 test cases. However, in your branch, `lit` is missing 1503 test cases.

dtcxzyw:
Why not keep these tests out-of-tree like SPEC benchmarks? I think they are licensed under GPL.

asb:
> Why not keep these tests out-of-tree like SPEC benchmarks? I think they are licensed under GPL.

The test suite already has a variety of different licenses, including GPL. I think it's convenient to have the sources in-tree when the license allows it. Are there any other tests that are kept out of tree that are open source?

mohammed-nurulhoque:
> Build failed on my end, please consider add the following

Fixed. The error flags where supposed to be excluded by regex

jerryzj:
> Do you mean gcc main? This PR only added the tests directly in gcc.dg without the subdirectories. If you mean the main of this repo, this PR only adds tests. It doesn't remove.

No. I meant llvm-test-suite.
Your patch causes the entire test suite to lose many test cases. Another thing I noticed is that your newly added tests don't even start. You can check the `test.log` for more details.

The test result of llvm-test-suite `main`

```
********************
Failed Tests (2):
  test-suite :: SingleSource/UnitTests/X86/x86-dyn_stack_alloc_realign.test
  test-suite :: SingleSource/UnitTests/X86/x86-dyn_stack_alloc_realign2.test


Testing Time: 152.57s
  Passed: 2509
  Failed:    2
```

On your branch

```
********************
Failed Tests (2):
  test-suite :: SingleSource/UnitTests/X86/x86-dyn_stack_alloc_realign.test
  test-suite :: SingleSource/UnitTests/X86/x86-dyn_stack_alloc_realign2.test


Testing Time: 151.32s
  Passed: 948
  Failed:   2
```

mohammed-nurulhoque:
> Your patch causes the entire test suite to lose many test cases. 

Oh that was probably because my branch wasn't rebased to latest master. I did now

> Another thing I noticed is that your newly added tests don't even start

It wasn't included by default. Had to be included explcitly with -DTEST_SUITE_SUBDIRS. Now changed to include by default.

mohammed-nurulhoque:
Gentle ping for this

mablinov:
Hi all, I've picked up and continued Mohammed's work in a new MR [here](https://github.com/llvm/llvm-test-suite/pull/72) - I hope this is the appropriate workflow, please let me know if I should do this differently.

PR #50, opened 2023-11-14
Port tools/timeit to windows (AArch64, x64)

omjavaid:
This patch implements timeit timing tool found at tools/timeit.c for Windows.
It is required for compile time and run time calculations by the llvm-testsuite.

At this moment llvm testsuite tests can not be compiled with clang-cl MSVC combination
however we have also made required modifications to the cmake files to enable building
timeit on windows with clang-cl/cl.

In developer console with MSVC following command can be used to successfully build this code:

set CC=c:\\work\\llvm-dev\\build\\bin\\clang-cl.exe
set CXX=c:\\work\\llvm-dev\\build\\bin\\clang-cl.exe

cmake -G Ninja -DTEST_SUITE_SUBDIRS= ..\\llvm-test-suite

Phabricator Review: https://reviews.llvm.org/D153263

### DIFF
--- a/cmake/modules/Host.cmake
+++ b/cmake/modules/Host.cmake
@@ -1,4 +1,10 @@
-set(TEST_SUITE_HOST_CC "cc" CACHE STRING "C compiler targetting the host")
+
+if(CMAKE_HOST_SYSTEM_NAME STREQUAL "Windows" AND NOT TEST_SUITE_HOST_CC)
+  set(TEST_SUITE_HOST_CC "clang" CACHE STRING "C compiler targetting the host")
+else()
+  set(TEST_SUITE_HOST_CC "cc" CACHE STRING "C compiler targetting the host")
+endif()
+
 mark_as_advanced(TEST_SUITE_HOST_CC)
 
 function(llvm_add_host_executable targetname exename)
@@ -21,6 +27,11 @@ function(llvm_add_host_executable targetname exename)
 
     list(APPEND _objs ${_objfile})
   endforeach ()
+
+  # Append ".exe" to the executable name on Windows
+  if(WIN32)
+    set(exename "${exename}.exe")
+  endif()
 
   add_custom_command(OUTPUT ${exename}
     COMMAND ${TEST_SUITE_HOST_CC} ${_objs}
--- a/tools/timeit.c
+++ b/tools/timeit.c
@@ -13,11 +13,16 @@
 
 #include <errno.h>
 #include <signal.h>
+
+#if _WIN32
+#include <windows.h>
+#else
 #include <unistd.h>
 
 #include <sys/resource.h>
 #include <sys/time.h>
 #include <sys/wait.h>
+#endif
 
 /* Enumeration for our exit status codes. */
 enum ExitCode {
@@ -52,7 +57,12 @@ static int g_posix_mode = 0;
 static int g_timeout_in_seconds = 0;
 
 /* \brief If non-zero, the PID of the process being monitored. */
+#ifndef _WIN32
 static pid_t g_monitored_pid = 0;
+#else
+static DWORD g_monitored_pid = 0;
+static HANDLE g_monitored_phandle = INVALID_HANDLE_VALUE;
+#endif
 
 /* \brief If non-zero, the path to attempt to chdir() to before executing the
  * target. */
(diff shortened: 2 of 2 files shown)


cmake/modules/Host.cmake
```diff
@@ -1,4 +1,10 @@
-set(TEST_SUITE_HOST_CC "cc" CACHE STRING "C compiler targetting the host")
+
+if(WIN32 AND NOT TEST_SUITE_HOST_CC)
```
mstorsjo, line 2: Just a note here; `WIN32` indicates whether a cross compiler targets windows, while we probably want to know what host we're running on. A more accurate condition might be to check `CMAKE_HOST_SYSTEM_NAME` if that's what you're intending here?

Anyway, I'm not a fan of hardcoded defaults, but I think this probably is ok (and it can be overridden by the user when necessary).

tools/timeit.c
```diff
@@ -74,57 +81,80 @@ static const char *g_target_redirect_stderr = 0;
...
+/* \brief If platform does not define pid_t type, then define it as int. */
+typedef int pid_t;
+#endif
+
 /* @name Resource Limit Variables */
 /* @{ */
 
 /* \brief If non-sentinel, the CPU time limit to set for the target. */
-static rlim_t g_target_cpu_limit = ~(rlim_t) 0;
+static rlim_t g_target_cpu_limit = ~(rlim_t)0;
```
mstorsjo, line 96: These formatting changes would probably be good to leave out of the commit.

tools/timeit.c
```diff
@@ -74,57 +81,80 @@ static const char *g_target_redirect_stderr = 0;
...
+
+#ifndef _WIN32
 static void terminate_handler(int signal) {
   /* If we are monitoring a process, kill its process group and assume we will
    * complete normally.
    */
   if (g_monitored_pid) {
-    fprintf(stderr, ("%s: error: received signal %d. "
-                     "killing monitored process(es): %s\n"),
+    fprintf(stderr,
```
mstorsjo, line 146: Unnecessary formatting change.

tools/timeit.c
```diff
@@ -74,57 +81,80 @@ static const char *g_target_redirect_stderr = 0;
...
             g_program_name, signal, g_target_program);
 
     /* Kill the process group of monitored_pid. */
     kill(-g_monitored_pid, SIGKILL);
     return;
   }
 
-  fprintf(stderr, "%s: error: received signal %d. exiting.\n",
-          g_program_name, signal);
+  fprintf(stderr, "%s: error: received signal %d. exiting.\n", g_program_name,
```
mstorsjo, line 156: Unnecessary formatting change.

tools/timeit.c
```diff
@@ -140,10 +170,92 @@ static void timeout_handler(int signal) {
...
+static FILE *windows_fopen(const char *filename, const char *mode) {
+  FILE *file;
+  if (fopen_s(&file, filename, mode))
+    return NULL;
+  return file;
+}
+#define fopen(filename, mode) windows_fopen(filename, mode)
+
+/* Process information struct global variable. */
+PROCESS_INFORMATION proc_info = {0};
```
mstorsjo, line 203: As other globals in the file seems to be using the naming convention of a `g_` prefix for globals, we could probably use that here as well. And it might be good to point out that the code generally uses local variables with `pid_t` for passing the reference to a process around, but here we're cheating with global variables instead, which only allows one single child process.

tools/timeit.c
```diff
@@ -219,7 +407,9 @@ static int monitor_child_process(pid_t pid, double start_time) {
     fprintf(fp, "%-10s %.4f\n", "sys", sys_time);
     fclose(fp);
   }
-
+#ifdef _WIN32
+  CloseHandle(h_output_file);
```
mstorsjo, line 411: Shouldn't we have an `if (h_output_file != INVALID_HANDLE_VALUE)` around this?

cmake/modules/Host.cmake
omjavaid: I have update this in new revision.

tools/timeit.c
omjavaid: fixed.

tools/timeit.c
omjavaid: proc_info is no longer a global after latest changes.

omjavaid: Also commented about single threaded behaviour of the app in recent changes.

tools/timeit.c
omjavaid: yes. I have fixed this.

mstorsjo (approved):
LGTM, assuming that you've tested it on both Unix and Windows.

PR #51, opened 2023-11-16
Fix CMake modules to allow build on windows

omjavaid:
This patch makes some minor adjustments to top level CMakeLists.txt and CopyDir, Testsuite modules to allow testsuite build on windows.


tarunprabhu (approved):
LGTM. But I can't test these out myself since I don't have access to a Windows machine.

PR #52, opened 2023-11-16
Add find compiler-rt libs cmake module for windows

omjavaid:
This patch adds find compiler-rt libs function imported from llvm-project/cmake/modules to testsuite repository. Some tests require to link compiler-rt builtins library on windows and this function helps locate the library path from install dir.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -237,6 +237,8 @@ if(CMAKE_SYSTEM_NAME STREQUAL "AIX")
   endif()
 endif()
 
+include(HandleCompilerRT)
+
 include(TestSuite)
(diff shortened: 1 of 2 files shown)


tarunprabhu (approved):
LGTM. Thanks!

petrhosek:
I missed this PR earlier, sorry about that. Why not directly copy https://github.com/llvm/llvm-project/blob/21edd381e403079baa2d01754e3d3ae7b3469932/cmake/Modules/HandleCompilerRT.cmake? That would have made it easier to keep the two modules in sync and I expect the support for finding compiler-rt on other platforms would be useful in test-suite as well.

PR #53, opened 2023-11-16
[Fortran/gfortran] Enable previously disabled tests

tarunprabhu:
Several tests that previous failed now pass because of more features having been implemented and several bugs having been fixed.

### DIFF
--- a/Fortran/gfortran/regression/gomp/DisabledFiles.cmake
+++ b/Fortran/gfortran/regression/gomp/DisabledFiles.cmake
@@ -12,31 +12,21 @@ set(UNSUPPORTED_FILES "")
 file(GLOB UNIMPLEMENTED_FILES CONFIGURE_DEPENDS
   # unimplemented: procedure pointers
   atomic-10.f90
-  declare-variant-19.f90
   proc_ptr_1.f90
-  proc_ptr_2.f90
 
   # unimplemented: procedure pointer arguments
-  pr44036-1.f90
   pr44036-2.f90
 
   # unimplemented: coarray in procedure interface
   declare-simd-coarray-lib.f90
 
   # unimplemented: gather rhs LEN parameters in assignment to allocatable
-  atomic-28.f90
-  defaultmap-2.f90
   defaultmap-3.f90
   defaultmap-4.f90
   defaultmap-5.f90
   defaultmap-6.f90
 
   # unimplemented: support for polymorphic types
-  associate2.f90
-  class-firstprivate-1.f90
-  class-firstprivate-2.f90
-  class-firstprivate-3.f90
-  class-firstprivate-4.f90
   omp_parallel_1.f90
   pr102621.f90
   pr52531.f90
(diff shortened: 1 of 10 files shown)


clementval (approved):
Thanks!

tarunprabhu:
Could someone test this on AArch64 please. I have tested it on an older machine, but it would be good to test it one something newer before I merge it in.

luporl (commented):
I've run the tests on a Ampere Altra Mt Jade and only 2 of them failed.
gdb just says they got a SIGSEGV on Fortran's main.

jeanPerier:
> I've run the tests on an Ampere Altra Mt Jade and only 2 of them failed.
> gdb just says they got a SIGSEGV on Fortran's main.

@tarunprabhu , these are `dg-do run` `dg-shouldfail` tests, you probably do not want to re-enable any such tests which most likely have undefined behaviors for runtime user errors we are not catching. There may be more than the two that @luporl found given the undefined behavior may just randomly segfault or not.

Fortran/gfortran/regression/gomp/DisabledFiles.cmake
```diff
@@ -45,84 +35,49 @@ file(GLOB UNIMPLEMENTED_FILES CONFIGURE_DEPENDS
 
   # unimplemented: derived type components with non default lower bounds
   depend-iterator-1.f90
-  depend-iterator-2.f90
-
-  # unimplemented: assumed rank in procedure interface
-  pr94672.f90
```
kiranchandramohan, line 51: Not sure how this started working.

kiranchandramohan:
I see the same results as @luporl on AArch64.
Failed Tests (2):
  test-suite :: Fortran/gfortran/regression/gfortran-regression-execute-regression__pointer_check_11_f90.test
  test-suite :: Fortran/gfortran/regression/gfortran-regression-execute-regression__pointer_check_12_f90.test

tblah:
I also see pointer_check_11.f90 and pointer_check_12.f90 fail on aarch64

tarunprabhu:
Thanks @luporl, @kiranchandramohan, @jeanPerier and @tblah for looking at this. I wrote a script to identify the tests that were passing and automatically enable them. Both that script and the gfortran test suite's build system needs more work as identified by @jeanPerier separately as well. I'll make a manual pass over it and remove any questionable tests in addition to those that you have identified. 

I'll work on the build system after this.

Fortran/gfortran/regression/gomp/DisabledFiles.cmake
tarunprabhu: This is expected to raise a compile-time error with the message 

> 'array' not specified in enclosing 'parallel'

In flang, this now errors with the message:

> The DEFAULT(NONE) clause requires that 'array' must be listed in a data-sharing attribute clause

Are these not equivalent? 

It was probably hitting a TODO earlier with the assumed rank error message.

tarunprabhu:
Abandoning this pull request. There are several issues with the test suite as is and I will fix those before enabling more tests.

PR #55, opened 2023-11-23
[Fortran] Comment out tests with missing semantic checks

jeanPerier:
These tests are looking for a semantic error that flang is not emitting but they used to pass because a TODO for procedure pointers was emitted in lowering. Lowering implemented procedure pointers, so these test now fail. I opened an issue against semantics, but I want to comment out these tests so that we can still land the lowering implementation of procedure pointers that is correct (the lowering change was reverted for now).

See https://github.com/llvm/llvm-project/issues/73215


kiranchandramohan (approved):
LGTM.

PR #56, opened 2023-11-28
[MicroBenchmark,LoopInterleaving] Re-land - Check performance impact of Loop Interleaving Count with varying loop iterations

nilanjana87:
This is for re-committing the microbenchmark that attempts to find the right loop trip count threshold for deciding whether to interleave a loop or not for different types of loops, such as loops with or without reduction inside it, loops with or without vectorization inside it. For context, the loop vectorizer uses a threshold of [TinyTripCountInterleaveThreshold](https://github.com/llvm/llvm-project/blob/56f7c7e52fe34fefda217b1194f681333159a809/llvm/lib/Transforms/Vectorize/LoopVectorize.cpp#L5672) that is currently set to 128.

The timeout issue in the previous PR https://github.com/llvm/llvm-test-suite/pull/26 is solved in this PR by using a pre-processor configuration flag of ALL_LOOP_IC_TESTS, which must be passed during compilation to run the whole set of tests, whereas by default it runs a reduced test set.


antmox (approved):
LGTM. This one should be fine IMO.
~250secs instead of 2500secs for the 1st version.

PR #60, opened 2023-12-08
[Fortran/gfortran] Add a simple not tool to the test suite

tarunprabhu:
Add a tool very similar to LLVM's not utility which inverts the return code of a command. The major difference is that the utility is not dependent on LLVM's libraries and performs less error checking than LLVM's not. It does support the
 --crash option when the command is expected to crash.

There is a need for something similar to the `not` utility in the gfortran test suite where some tests are [expected to crash](https://github.com/llvm/llvm-test-suite/pull/57#discussion_r1420934396). However, we cannot rely on the utility being available since the LLVM build directory may not be available when running the test suite. Unless LLVM was built with `-DLLVM_INSTALL_UTILS=On`, the utilities will not be available even if clang/flang are present.

In keeping with some of the other tools in the test suite which were borrowed from LLVM, this has been stripped down and made independent of LLVM's libraries.

### DIFF
--- a/tools/not.cpp
+++ b/tools/not.cpp
@@ -1,0 +1,106 @@
+//===- not.cpp - The 'not' testing tool -----------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+// Usage:
+//   not cmd
+//     Will return true if cmd doesn't crash and returns false.
+//   not --crash cmd
+//     Will return true if cmd crashes (e.g. for testing crash reporting).
+
+// This file is a stripped down version of not.cpp from llvm/utils. This does
+// not depend on any LLVM library.
+
+#include <cstdlib>
+#include <iostream>
+#include <sstream>
+
+#ifdef _WIN32
+#define WIN32_LEAN_AND_MEAN
+#define NOMINMAX
+#include <windows.h>
+#endif
+
+int main(int argc, const char **argv) {
+  bool expectCrash = false;
+
+  ++argv;
+  --argc;
+
+  if (argc > 0 && std::string(argv[0]) == "--crash") {
+    ++argv;
+    --argc;
+    expectCrash = true;
+
+    // Crash is expected, so disable crash report and symbolization to reduce
+    // output and avoid potentially slow symbolization.
+#ifdef _WIN32
+    SetEnvironmentVariableA("LLVM_DISABLE_CRASH_REPORT", "1");
+    SetEnvironmentVariableA("LLVM_DISABLE_SYMBOLIZATION", "1");
+#else
+    setenv("LLVM_DISABLE_CRASH_REPORT", "1", 0);
+    setenv("LLVM_DISABLE_SYMBOLIZATION", "1", 0);
+#endif
+  }
+
+  if (argc == 0)
+    return 1;
+
+  std::stringstream ss;
+  ss << argv[0];
+  for (int i = 1; i < argc; ++i)
+    ss << " " << argv[i];
+  std::string cmd = ss.str();
+
+  int result = std::system(cmd.c_str());
+  int retcode = 0;
+  int signal = 0;
+
+#ifdef _WIN32
+  // Handle abort() in msvcrt -- It has exit code as 3.  abort(), aka
+  // unreachable, should be recognized as a crash.  However, some binaries use
+  // exit code 3 on non-crash failure paths, so only do this if we expect a
+  // crash.
+  if (expectCrash && result == 3) {
+    retcode = 3;
+    signal = 1;
+  } else if (errno) {
+    // If the command interpreter was not found, errno will be set and 0 will
+    // be returned. It is unlikely that this will happen in our use case, but
+    // check anyway.
+    retcode = 1;
+    signal = 1;
+  } else {
+    // On Windows, result is the exit code, except for the special case above.
+    retcode = result;
+    signal = 0;
+  }
+#elif defined(WEXITSTATUS) && defined(WTERMSIG)
+  // On POSIX systems and Solaris, result is a composite value of the exit code
+  // and, potentially, the signal that caused termination of the command.
+  retcode = WEXITSTATUS(result);
+  signal = WTERMSIG(result);
+#else
+#error "Unsupported system"
+#endif
+
+  // If signal is non-zero, the command caused a crash, usually SIGABRT.
+  if (signal) {
+    if (expectCrash)
+      return EXIT_SUCCESS;
+    return EXIT_FAILURE;
+  }
+
+  // The command exited normally. If the command was expected to crash, return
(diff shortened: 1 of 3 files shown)


cabreraam:
Thanks for working on this! I'll have a look next week.

tools/not.cpp
```diff
@@ -0,0 +1,76 @@
...
+  if (result < 0) {
+    if (expectCrash)
+      return 0;
+    return 1;
+  }
+
+  if (expectCrash)
+    return 1;
+
+  return result == 0;
```
tblah, line 75: Doesn't 0 mean successful termination on unix-like systems? In which case, wouldn't this not invert the return status of the program?

Maybe this would be clearer if you used the EXIT_SUCCESS constant from https://en.cppreference.com/w/c/program/EXIT_status

tblah: Oh I understand now - when result is 0, the condition evaluates to true (1) and so this inverts the error code. Nevermind.

tarunprabhu: I started to respond and then got stuck thinking more closely about what this code is doing, and there are some issues here too. The original code uses `llvm::sys::ExecuteAndWait` which returns -2 on an abort or a timeout. The `result < 0` condition checks this. But I am not sure if that is the case with the libc `system` call that I have used. I will try to find out, at least on Linux, and add comments to the code accordingly.

Thanks for raising the question :)

yi-wu-arm: system does not directly returns exit_code, at least not on Linux. (It returns a status code and be then interpreted as exit snd stopsig code) There is a macro can be used to get exit code from status: https://en.cppreference.com/w/cpp/utility/program/system
On POSIX systems, the return value can be decomposed using [WEXITSTATUS and WSTOPSIG](https://pubs.opengroup.org/onlinepubs/9699919799/functions/wait.html).
If I remember correctly the WEXITSTATUS calculate exit code by and 0xFF with a 8 shift left? 
This does not apply on Windows, I am not sure if the return code from std::system matches the Windows error code or not https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes--0-499-
TL;DR: so basically a value of 0 means no problem on Windows and Linux

tarunprabhu: Thanks @tblah and @yi-wu-arm. 

I have fixed the code to check for a signal that causes termination and added some comments explaining what is going on.

tblah (approved):
LGTM, thanks!

tools/not.cpp
```diff
@@ -63,14 +64,23 @@ int main(int argc, const char **argv) {
     result = -3;
 #endif
 
-  if (result < 0) {
+  // On POSIX systems, result is a composite value of the exit status and,
+  // potentially, the signal that caused termination of the command.
+  int retcode = WEXITSTATUS(result);
+  int signal = WTERMSIG(result);
+
```
yi-wu-arm, line 71: You probably don't have access to these function on Windows, what about an 
```cpp
# ifdef _Windows
//Windows stuff
#else
// Probably not the best way to handle POSIX
WEXITSTATUS
WTERMSIG
#endif
```

tools/not.cpp
```diff
@@ -0,0 +1,86 @@
...
+// This file is a stripped down version of not.cpp from llvm/utils. This does
+// not depend on any LLVM library.
+
+#include <cstdlib>
+#include <iostream>
+#include <sstream>
+
+#ifdef _WIN32
+#include <windows.h>
```
yi-wu-arm, line 22: nits: add ... before `#include <windows.h> to reduce the size of the Win32 header files included.
```cpp
#define WIN32_LEAN_AND_MEAN
#define NOMINMAX
#include <windows.h>
```

tools/not.cpp
```diff
@@ -0,0 +1,86 @@
...
+  if (argc == 0)
+    return 1;
+
+  std::stringstream ss;
+  ss << argv[0];
+  for (int i = 1; i < argc; ++i)
+    ss << " " << argv[i];
+  std::string cmd = ss.str();
+
+  int result = std::system(cmd.c_str());
```
cabreraam, line 56: I'm not sure if I'm reading this correctly, but if you include `argv[0]` -- which should be your `not` utility in this case -- in the `ss` variable, wouldn't the `std::system(cmd.c_str())` just recursively call
```bash
./not [--crash] <cmd>
```
recursively? Which would just run `not` in an infinite loop?
Again, not sure if I'm reading this correctly.

cabreraam (commented):
I can't speak to any of the Windows stuff, but it mostly looks good to me! I just had one question about how you invoke `std::system`.

tools/not.cpp
tarunprabhu: > You probably don't have access to these functions on Windows, what about a structure similar to this:
> 
> ```c++
> #ifdef _Win32
> //Windows stuff
> #else
> // Probably not the best way to check POSIX
> WEXITSTATUS
> WTERMSIG
> #endif
> ```

For Windows, is reasonable to assume that if the return code < 0, then the command aborted? This seems to suggest that it is: [https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/system-wsystem?view=msvc-170](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/system-wsystem?view=msvc-170).

Why do you have a comment saying, "Probably not the best way to check POSIX"?

tools/not.cpp
tarunprabhu: Done

tools/not.cpp
yi-wu-arm: I am not an expert on Windows stuff but based on the doc, I think that it means if there is problem with command-interpreter (cmd in this case), then it will return -1 and set errno. If cmd is fine and the command has problem, the return value will be cmd's return value.
e.g. if you test it with `int val = std::system("InvalidCommand");` val will be set to 0. My best guess in this case is to set retcode to the return value of `std::system()`, and if `returnVal==-1` then you set signal to errno.

Just because it isn't Windows doesn't mean it is POSIX-compliant? You can `#if defined(WIFSTOPPED) && defined(WSTOPSIG)` and then get exitstatus and sig, if it isn't defined then just take whatever value from `std::system`. I guess you can just leave it as is, unless we need to support system that is not Windows nor POSIX. because each of them may have their own implementation and ask you to match all the cases would be impossible.

tarunprabhu: That's a fair point. I have added checks for those macros, checked explicitly for `errno` on Windows and for the `WEXITSTATUS` and `WTERMSIG` macros. 

I can only think of Solaris as a non-POSIX system other than Windows that is likely to be of interest. It seems as if those macros [are defined](https://docs.oracle.com/cd/E36784_01/html/E36873/stdlib-3head.html) there. 

On any other system, I force a build error. I think it is safer than just using the result of `std::system` in such cases, but I can use a different strategy if folks feel differently.

cabreraam (approved):
I will defer to @yi-wu-arm

Otherwise, LGTM

yi-wu-arm (approved):
Thanks for addressing all my comment! I may have over-complicate this stuff.

tools/not.cpp
yi-wu-arm: LGTM, thanks!

PR #62, opened 2023-12-11
[flang] Scrub lists of failing tests in DisabledFiles.cmake

klausler:
Remove many tests from Fortran/gfortran/regression/DisabledFiles.cmake that now pass.  Move others around from more general classifications to groups with specific diagnoses like "unimplemented feature" or "bad test".


tblah (approved):
Thanks for this. These are passing for me on aarch64

klausler:
Tarun, any comment?

tarunprabhu:
> Tarun, any comment?

Sorry for the delay, got tangled up with my other projects. I'll take this for a spin this afternoon.

tarunprabhu:
Thanks for the detailed classification. I think more tests will be removed from this list once the build system is fixed and the `shouldfail` tests are handled correctly.

tarunprabhu (approved):

PR #63, opened 2023-12-14
Additional Redlist/Greenlist functionality to External/smoke and External/HeCBench

AaronLiu20:
creating a green list and redlist containing tests that fail at compilation and a separate list for tests that fail at runtime


ajarmusch:
@AaronLiu20 thanks for adding the description, LGTM

PR #66, opened 2023-12-18
Fixing External/HecBench

AaronLiu20:
finishing the conversion to work with new system of declaring which list  to use


ajarmusch (approved):
LGTM

PR #68, opened 2023-12-20
litsupport changes for windows support

omjavaid:
This patch makes changes required in litsupport/* to support running llvm testsuite on windows platform.

### DIFF
--- a/litsupport/modules/timeit.py
+++ b/litsupport/modules/timeit.py
@@ -5,7 +5,7 @@ import re
 
 
 def _mutateCommandLine(context, commandline):
-    timefile = context.tmpBase + ".time"
+    timefile = os.path.normpath(context.tmpBase + ".time")
     config = context.config
     cmd = shellcommand.parse(commandline)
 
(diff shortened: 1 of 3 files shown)


litsupport/shellcommand.py
```diff
@@ -88,6 +88,7 @@ def parse(commandline):
     and will throw an exception if the commandline uses unsupported features.
     """
     result = ShellCommand()
+    commandline = commandline.replace("\\", "/") if os.name == 'nt' else commandline
```
tarunprabhu, line 91: Could you just use the if statement syntax here instead of an if expression. 

```
if os.name == 'nt':
    commandline = commandline.replace("\\", "/")
```

litsupport/modules/timeit.py
```diff
@@ -42,17 +46,17 @@ def _mutateCommandLine(context, commandline):
...
     stdin = cmd.stdin
     if stdin is not None:
         if not os.path.isabs(stdin) and workdir is not None:
-            stdin = os.path.join(workdir, stdin)
+            stdin = os.path.join(workdir, os.path.normpath(stdin))
         args += ["--redirect-input", stdin]
         cmd.stdin = None
     else:
-        args += ["--redirect-input", "/dev/null"]
+        args += ["--redirect-input", "/dev/null"] if os.name != 'nt' else []
```
tarunprabhu, line 59: Could you just write this using the if statement syntax instead of using an if expression. i.e.

```
if os.name != 'nt':
    args += ["--redirect-input", "/dev/null"]
```

tarunprabhu (changes requested):
Thanks for working on this. It looks fine except for the if expressions.

litsupport/shellcommand.py
omjavaid: Fixed by latest commit.

litsupport/modules/timeit.py
omjavaid: likewise.

tarunprabhu (approved):
Thanks for making the changes.

PR #69, opened 2023-12-20
Fix Fortran tests for Windows platform

omjavaid:
This patch makes minor changes to ignore windows line endings when Fortran tests are run on windows platform.

### DIFF
--- a/Fortran/UnitTests/fcvs21_f95/CMakeLists.txt
+++ b/Fortran/UnitTests/fcvs21_f95/CMakeLists.txt
@@ -82,7 +82,8 @@ endif()
 # Skip some tests that currently fail with flang on AArch64 when using libpgmath.
 # FIXME: Reenable these tests after libpgmath is fixed or after we no longer
 # depend on it.
-if (CMAKE_Fortran_COMPILER_ID STREQUAL "LLVMFlang" AND ARCH STREQUAL "AArch64")
+if (CMAKE_Fortran_COMPILER_ID STREQUAL "LLVMFlang"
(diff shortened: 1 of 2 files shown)


Fortran/UnitTests/fcvs21_f95/CMakeLists.txt
```diff
@@ -82,7 +82,8 @@ endif()
 # Skip some tests that currently fail with flang on AArch64 when using libpgmath.
 # FIXME: Reenable these tests after libpgmath is fixed or after we no longer
 # depend on it.
-if (CMAKE_Fortran_COMPILER_ID STREQUAL "LLVMFlang" AND ARCH STREQUAL "AArch64")
+if (CMAKE_Fortran_COMPILER_ID STREQUAL "LLVMFlang"
+    AND (ARCH STREQUAL "AArch64" OR TARGET_OS STREQUAL "Windows"))
```
tarunprabhu, line 86: If I understand this correctly, the tests will be disabled on Windows on all hardware platforms and will be disabled on all OS'es on AArch64. I just want to double-check that this is what you intended?

tarunprabhu:
Apologies @omjavaid, I somehow didn't see that you had requested a review from me.

Fortran/UnitTests/fcvs21_f95/CMakeLists.txt
omjavaid: Yes. This fails on windows regardless of the underlying architecture. I am currently trying to enable everything which works.

tarunprabhu (approved):
LGTM. Thanks for this.

PR #70, opened 2023-12-21
[gcc-c-torture] use -fwrapv for 950704-1.c (signed overflow)

antmox:
test fails since #76044


nikic (approved):

fhahn:
Thanks for fixing this!

PR #71, opened 2024-01-02
[flang] Remove more working/fixed tests, classify others

klausler:
Fixes have gone into llvm-project/main that fix several tests in Fortran/gfortran/regression/DisabledFiles.cmake; delete them from this file.  Other tests were noticed to be working, and were also deleted.  Some others were moved from generically described blocks of tests into blocks with more specific descriptions.


tarunprabhu (approved):
LGTM. 

Perhaps someone could check on AArch64 as well.

tblah (approved):
Passes on arch64 with `-O3`. Thanks for this!

PR #73, opened 2024-01-09
[tools] Fix not utility to work on iOS

tarunprabhu:
This addresses issue #77137 (std::system is not available on iOS). The implementation uses posix_spawn on iOS but std::system elsewhere.

This has only been tested on Linux on X86 and an older AArch64. 

@jroelofs, could you try this on an iOS and let me know if it works there?

### DIFF
--- a/tools/not.cpp
+++ b/tools/not.cpp
@@ -17,8 +17,6 @@
 #include <cstdlib>
 #include <iostream>
 #include <sstream>
-#include <spawn.h>
-#include <wait.h>
 
 #ifdef _WIN32
 #define WIN32_LEAN_AND_MEAN
@@ -27,6 +25,8 @@
 #endif
 
 #ifdef __APPLE__
+#include <spawn.h>
+#include <wait.h>
 #include <TargetConditionals.h>
 #endif
 
@@ -64,9 +64,8 @@ int main(int argc, char* const* argv) {
   std::string cmd = ss.str();
   result = std::system(cmd.c_str());
 #else
-  char* const* environ = NULL;
   pid_t pid;
-  if (posix_spawn(&pid, argv[0], NULL, NULL, argv, environ))
+  if (posix_spawn(&pid, argv[0], NULL, NULL, argv, NULL))
     return EXIT_FAILURE;
   if (waitpid(pid, &result, WUNTRACED | WCONTINUED) == -1)
     return EXIT_FAILURE;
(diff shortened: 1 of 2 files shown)


tools/not.cpp
```diff
@@ -17,14 +17,20 @@
 #include <cstdlib>
 #include <iostream>
 #include <sstream>
+#include <spawn.h>
```
yi-wu-arm, line 20: If I remember correctly `spawn.h` and is not a standard library, and not supported on Windows? you may not have access to `wait.h` on Windows as well... I saw you used `waitpid` only on IOS ifdef, so could you move it under `ifdef __APPLE__` as well?

tblah (commented):
Changes look sensible to me, assuming that posix_spawn and waitpid on iOS are standards compliant. However, I have no experience writing software for iOS. Please wait for approval by somebody who can test on iOS.

tools/not.cpp
```diff
@@ -49,13 +55,22 @@ int main(int argc, const char **argv) {
...
   for (int i = 1; i < argc; ++i)
     ss << " " << argv[i];
   std::string cmd = ss.str();
-
-  int result = std::system(cmd.c_str());
+  result = std::system(cmd.c_str());
+#else
+  char* const* environ = NULL;
+  pid_t pid;
+  if (posix_spawn(&pid, argv[0], NULL, NULL, argv, environ))
```
yi-wu-arm, line 69: could you put a NULL instead of an environ?

tblah:
As posix_spawn is part of POSIX, it is probably reasonably well available on other platforms (although I don't know about Windows). Perhaps we could get away without the #ifdef here. I don't feel strongly about this.

yi-wu-arm (commented):
Thanks for fixing this! I am not very familiar with iOS process calls.

tools/not.cpp
tarunprabhu: Done

tools/not.cpp
tarunprabhu: Done

tools/not.cpp
```diff
@@ -24,7 +24,13 @@
 #include <windows.h>
 #endif
 
-int main(int argc, const char **argv) {
+#ifdef __APPLE__
+#include <spawn.h>
+#include <wait.h>
```
jroelofs, line 29:
```suggestion
#include <sys/wait.h>
```

jroelofs (approved):
LGTM with an include path fix.

jroelofs:
thank you!

PR #75, opened 2024-01-12
Remove signed integer overflow from test

nmosier:
Remove signed integer overflow from SingleSource/Regression/C/gcc-c-torture/execute/950704-1.c, which invoked undefined behavior. clang at HEAD just compiles this program to a single abort statement (w/ -O3 flag).


fhahn:
d10245e added `fwrapv` to the flags for this test, so the overflow shouldn't be an issue, while preserving the original test. Is this change still needed?

PR #77, opened 2024-01-19
[test-suite][Fortran] Ignore -fstrict-aliasing and -fschedule-insns

luporl:
Ignore these flags as they are currently not supported by flang.
Also re-enable pr43808.f90 test, that passes with this change.


jeanPerier (approved):
Makes sense

kkwli (approved):
Thanks for fixing it. LG

tarunprabhu (approved):
LGTM. Thanks.

PR #78, opened 2024-01-19
[SingleSource/Atomic] Add preliminary tests for atomic builtins.

Logikable:
There exist atomic IR unit tests and libatomic unit tests, but neither can test the atomicity and interoperability of atomic builtins and compiler-rt's atomic library. These tests aim to approximate behaviour encountered in user code.

These tests have caught issues in Clang. See
https://github.com/llvm/llvm-project/pull/74349 and https://github.com/llvm/llvm-project/pull/73176 for LLVM changes inspired by these tests.


SingleSource/UnitTests/Atomic/float_test.cpp
```diff
@@ -0,0 +1,135 @@
...
+void test_float_scalar_xchg() {
+  static constexpr T val = V >> right_shift<T>();
+  static constexpr T expected = val * kExpected;
+  std::vector<std::thread> pool;
+
+  for (int model : atomic_exchange_models) {
+    T afloat = 0;
+    T ffloat = 0;
+    for (int n = 0; n < kThreads; ++n) {
+      pool.emplace_back(looper_numeric_xchg_atomic<T>, &afloat, model);
```
MaskRay, line 65: Nit: llvm style https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements prefer to drop the braces here.

IIRC Google style clang-format often places the body on the same line, thus many people prefer `{}` to work around it.

SingleSource/UnitTests/Atomic/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+set(FP_TOLERANCE 1000000)
+execute_process(COMMAND ${CMAKE_C_COMPILER} --print-file-name=libclang_rt.atomic.so
+                OUTPUT_VARIABLE _path_to_libatomic
+                OUTPUT_STRIP_TRAILING_WHITESPACE)
+get_filename_component(_libatomic_dir ${_path_to_libatomic} DIRECTORY)
+set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${_path_to_libatomic} -Wl,-rpath=${_libatomic_dir}")
```
Meinersbur, line 6: Changing global variables is usually not a good idea. Prefer [`target_link_options`](https://cmake.org/cmake/help/latest/command/target_link_options.html), or at the very least or at least [add_link_options](https://cmake.org/cmake/help/latest/command/add_link_options.html) so it applies to all targets in this CMakeLists and subdirectories.

Please add a comment on why it is needed.

SingleSource/UnitTests/Atomic/big_test.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+#include <cstdio>
+#include <iostream>
+#include <thread>
+#include <vector>
+
+#include "util.h"
+
+static constexpr int kBigSize = 10;
+struct big {
```
Meinersbur, line 53: For readability, I suggest to call the type `big_t`.

SingleSource/UnitTests/Atomic/big_test.cpp
```diff
@@ -0,0 +1,122 @@
...
+                        int fail_model) {
+  for (int n = 0; n < kIterations; ++n) {
+    big desired, expected = {};
+    do {
+      desired = expected;
+      for (int k = 0; k < kBigSize; ++k) {
+        desired.v[k]++;
+      }
+    } while (!__atomic_compare_exchange(abig, &expected, &desired, true,
+                                        success_model, fail_model));
```
Meinersbur, line 69: I am a bit surprised `__atomic_compare_exchange` allows non-constant for `success_model` and `fail_model` since these may map to different instructions. Does Clang/gcc emit branches for them? Or does it possibly pessimize to the lowest common denominator in which case this would test just sequential consistency (even if `atomic_compare_exchange_models` contains other values)? Could be resolved by making them template arguments.

SingleSource/UnitTests/Atomic/float_test.cpp
```diff
@@ -0,0 +1,135 @@
...
+                << "nonatomic: " << ffloat << "\n";
+      if (lt(afloat, ffloat) || afloat < expected * (1 - kEpsilon) ||
+          afloat > expected * (1 + kEpsilon)) {
+        fail();
+      }
+    }
+  }
+}
+
+void test_float() {
```
Meinersbur, line 119:
```suggestion
void test_floating_point() {
```
Since one of the types is `float`, `test_float` implies it would only test single precision.

Meinersbur (commented):
Have you considered using GTest for these? It's already contained in the llvm-test-suite repository.

SingleSource/UnitTests/Atomic/big_test.cpp
Logikable: Clang emits branches. I'm not sure whether gcc emits branches, but it doesn't at least for int32s on x86.

Logikable:
Updated with suggestions.

> Have you considered using GTest for these? It's already contained in the llvm-test-suite repository.

The first draft of these tests was written with GTest, but I dropped it along the way. What are some reasons switching to GTest is worth the investment?

Meinersbur:
> The first draft of these tests was written with GTest, but I dropped it along the way. What are some reasons switching to GTest is worth the investment?

 * LIT already has infrastructure for GTest
 * cmpxchg/xchg verification should be exact (no floating-point rounding). GTest allows you to later add `ASSERT_NEAR` tests for atomic add/inc/mul etc without weakening the xchg tests.
 * Fails already at `ASSERT_NEAR` in the test, instead of fpcmp after the program has already finished.
 * fpcmp is textual comparison, floating-point epsilon-comparision is a workaround on top of it.

MaskRay:
@tstellar @zeroomega It seems that llvm-test-suite will need a `third-party/unittest/googletest/` as well.

pogo59:
>  llvm-test-suit already has a copy of googletest

Aha! Thank you. That is buried pretty deep.

> Maybe move it to a path like third_party/googletest and make it shared by all llvm-test-suite components?

SGTM.

Logikable:
To add some context to @MaskRay's comment, I have tried and failed to move the currently imported GoogleTest (at `Microbenchmarks/libs/benchmark/googletest` to a more centralized location so it can be used by any test in this suite.

I have a commit on the way that migrates my tests to GoogleTest, but in a pretty hackish way. Improvement suggestions are welcome.

Logikable:
I've put in the better part of a week into setting up the infrastructure for GoogleTest to work across the repository (instead of just within `Microbenchmarks/libs/benchmark`), and despite bodging numerous things to get it to work, it still doesn't. This in addition to the comments above, I think this will be a more significant effort than I can (or should) manage for this PR.

You can find my attempt at https://github.com/Logikable/llvm-test-suite/tree/bodge.

@Meinersbur can you provide a code example of what I can do to hook things up?

pogo59:
It might be informative to look at the llvm-project patch to move googletest to a new location:
https://github.com/llvm/llvm-project/commit/a11cd0d94ed3cabf0998a0289aead05da94c86eb

Logikable:
Thanks for sharing that patch. I want to be more firm about pushing back against me moving these tests to GTest for now. Getting GTest to work requires much more than moving it to a new location:

- When I build it by including its CMake file, its compile time is measured by timeit thanks to hhttps://github.com/llvm/llvm-test-suite/blob/main/CMakeLists.txt#L325-L328. But sometimes, timeit won't have been built by this point. I don't understand CMake or how the lists have been set up remotely enough to know how to fix this race condition. Do I specify a dependency, or do I somehow build GoogleTest before this instrumentation is added?
- There's a lot of infrastructure to get `llvm_singlesource()` and `llvm_multisource()` to work. Ideally, GTest has similar infrastructure, to avoid things like `target_link_libraries(${name} GTest::gtest_main)` in multiple places. I don't know that I'm the person to do this; it will take me weeks I don't have.

I would be happy to open a separate PR with my patch linked above once GTest has been successfully integrated into this repo.

pogo59:
It looks like you've made a real effort, and I can sympathize with the difficulties of recoding a bunch of CMake stuff.  At this point I'd think the best way forward is to go ahead with the non-GTest based tests, and file an issue to get gtest moved.

Meinersbur:
@Logikable My original question was just whether you considered it. Your code looked too GTest-like to just ignore it. However, I see your reasons and would be ok to move forward without it. May put a TODO in the source, for instance where FP_TOLERANCE is being worked around.

One more concern though: is explicit specifying the path to `libclang_rt.atomic.so` really necessary? Shouldn't the compiler find it itelf? It would be nice if we could keep the test-suite compiler-independent so we can check consistency with other compilers and/or using the system's  libatomic -- not everone always builds LLVM with compiler-rt.

MaskRay:
There is no option to instruct Clang driver to pass `libclang_rt.atomic.{a,so}` to ld, so `--print-file-name=libclang_rt.atomic.so` is perhaps the best we can do. I agree that having a way to test `libatomic` will be really useful. Perhaps add a CMake option?

arichardson:
For the tests that explicitly link against libclang_rt.atomic.so, would it make more sense to have them as tests in the compiler-rt directory? IIRC there are already some basic libatomic tests in that directory, so this would be a natural place to add more?

pogo59:
> For the tests that explicitly link against libclang_rt.atomic.so, would it make more sense to have them as tests in the compiler-rt directory?

@arichardson if you're referring to the llvm-project/compiler-rt directory, the point is that those are not executable tests. This one would actually be executed.

Logikable:
AFAICT, my tests are more expensive than the existing ones in atomic_test.c, taking O(minutes) to run. I'm not familiar with the test suite in llvm-project -- given this, is it okay to move them there?

pogo59:
One of the purposes of llvm-test-suite is to be a place to put longer-running tests. I would not put multiple-minute tests in llvm-project.

arichardson:
> AFAICT, my tests are more expensive than the existing ones in atomic_test.c, taking O(minutes) to run. I'm not familiar with the test suite in llvm-project -- given this, is it okay to move them there?

Ah I did not realize the tests would take that long to run. I agree that tests in llvm-project should not take long to run and something well above 1 minute seems problematic (I locally set the lit timeout to 120s per test to catch very slow running tests). Slow running tests have been quite annoying in the past when trying to run them on QEMU (e.g. the libc++ threading tests), so keeping them external makes sense. We could of course reduce the number of iterations but that reduces the probability of catching a race.

SingleSource/UnitTests/Atomic/big_test.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+void test_big_cmpxchg() {
+  std::vector<std::thread> pool;
+
+  for (int success_model : atomic_compare_exchange_models) {
+    for (int fail_model : atomic_compare_exchange_models) {
+      big_t abig = {};
+      big_t bbig = {};
+      for (int n = 0; n < kThreads; ++n)
+        pool.emplace_back(looper_big_cmpxchg, &abig, std::ref(bbig),
```
arichardson, line 82: Is there a specific reason to pass one as a pointer and one as a reference?

SingleSource/UnitTests/Atomic/big_test.reference_output
```diff
@@ -0,0 +1,6 @@
+10 threads; 1000000 iterations each; total of 10000000
+Testing big
+CMPXCHG: atomic: 10000000 10000000 10000000 10000000 10000000 10000000 10000000 10000000 10000000 10000000 
+      nonatomic: 9970397 9970397 9970397 9970397 9972269 9972269 9972269 9972269 9941180 9941180
```
arichardson, line 4: I'm not sure there is much value in including the non-atomic output in the reference (it could even match the atomic output in some cases, e.g. on a single CPU emulator). All we really care about in this test is that the atomic output is correct, right?

Would printing the non-atomic output to stderr instead of stdout keep the debugging feature while allowing us to drop the FP_TOLERANCE hack?

arichardson: Also thinking about this more, doesn't the FP_TOLERANCE comments also imply that the test will pass even if the values printed are wrong? So it would always require manual inspection?

SingleSource/UnitTests/Atomic/big_test.cpp
Logikable: Not anymore, I think this is a relic of an older implementation. I'll turn them all to pointers.

SingleSource/UnitTests/Atomic/big_test.reference_output
Logikable: It's a good idea to move them to stderr. I'll do that.

The tests themselves check for correctness (`if (aint < expected) printf('fail')`), so the matcher will fail too.

Logikable: What is "traditional_output" in this context? https://github.com/llvm/llvm-test-suite/blob/main/litsupport/modules/timeit.py#L30

I'm having trouble separating stdout from stderr, since this timeit lit module doesn't support it & redirecting to test.out simultaneously. I think it would hamper test readability if I wrote my own debug logging, so I suppose we could drop all numerical logging. Is anyone against this?

Logikable:
Updated with suggestions.

SingleSource/UnitTests/Atomic/float_test.cpp
```diff
@@ -0,0 +1,118 @@
...
+// result, although it is still outputted to aid in debugging.
+//
+// Each test is performed on all relevant types.
+//
+//===----------------------------------------------------------------------===//
+//
+// This file tests atomic operations on floating point types with aligned
+// memory addresses.
+//
+// The types tested are: float, double.
```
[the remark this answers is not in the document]
Logikable: Good point, I'll add it. It was originally omitted when this test was written with macros instead of templates.

SingleSource/UnitTests/Atomic/big_test.cpp
```diff
@@ -0,0 +1,103 @@
...
+//
+// For example, a looper that increments the shared pointer is expected to end
+// up with a value of 10^7. If its final value is not that, the test fails.
+//
+// Each test also tests the corresponding nonatomic operation with a separate
+// shared variable. Ideally, this value differs from the atomic "correct" value,
+// and the test can compare the two. In reality, some simpler operations (e.g.
+// those conducted through the ALU) can still end up with the correct answer,
+// even when performed nonatomically. These tests do not check the nonatomic
+// result, although it is still outputted to aid in debugging.
```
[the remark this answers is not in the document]
Logikable: Oops, this comment is not accurate anymore. We no longer output its results. Additionally, some of the tests do check the nonatomic result, which will likely cause flakiness down the line. I'll remove those checks.

The nonatomic version verifies that the test is meaningful in testing atomicity -- a test may appear to test atomicity, but pass even with nonatomic operations.

I feel the best way to move forward is to remove the nonatomic versions, and leave a note to future test writers to check that their test fails with nonatomic operations.

SingleSource/UnitTests/Atomic/big_test.cpp
```diff
@@ -0,0 +1,103 @@
...
+//
+// Each test is performed on all relevant types.
+//
+//===----------------------------------------------------------------------===//
+//
+// This file tests atomic operations on big objects with aligned memory
+// addresses.
+//
+// The types tested are: bigs.
+// The ops tested are: xchg, cmpxchg.
```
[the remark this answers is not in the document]
Logikable: Load/store tests were too large to fit in the margin of this test suite :)

I have yet to come up with a load/store test that verifies atomicity. I'll add a TODO.

SingleSource/UnitTests/Atomic/big_test.reference_output
Logikable: I've removed test result/numerical logging.

Logikable:
Updated with suggestions.

SingleSource/UnitTests/Atomic/util.h
```diff
@@ -0,0 +1,66 @@
...
+
+#include <cstdint>
+#include <cstdlib>
+#include <iostream>
+
+#define TEST16 1
+
+static constexpr uint64_t V = 0x4243444546474849;
+static constexpr int kThreads = 10;
+static constexpr int kIterations = 1'000'000;
```
arichardson, line 20: A lot of other tests/benchmarks use the CMake variable `SMALL_PROBLEM_SIZE` to reduce the time it takes to run the test. Since you said these take a long time to run, could this variable be set to a smaller number so that it's still possible to run them on an emulator or slow FPGA? Another option could be to pass the number of iterations and threads as a command line argument which would make it easy to change the number of iterations dynamically.

Logikable: We can do that. FWIW, on x64 with AVX2, even 10^6 iterations is (rarely) flaky for aligned integer ALU operations.

Given that the speed of each operation affects contention rates affects flakiness of the test, it could make sense to have a different constant for each file. That is, since slow tests (like misaligned integers) can experience contention even with 10^5 iterations, we can speed up the test suite by setting the number of iterations to be inversely proportional to the speed of each operation.

Logikable:
Oops. I git reset the wrong directory... remaking this PR now...

Logikable:
Created https://github.com/llvm/llvm-test-suite/pull/94.

PR #79, opened 2024-01-24
[cmake][LoongArch] Add testsuite framework for SX/ASX

leecheechen:


leecheechen:
cc @SixWeining @yjijd

cmake/modules/DetectLoongArchSupportVector.cmake
```diff
@@ -0,0 +1,47 @@
+##===- DetectLoongArchSupportVector.cmake ---------------------------------===##
+#
+# Performs a try_run to determine if the target machine can run the vector
```
SixWeining, line 3:
```suggestion
# Performs a try_run with a simple program calling LSX/LASX builtin to determine if the LoongArch CPU supports vector instructions.
```

cmake/modules/DetectLoongArchFeatureASX.c
```diff
@@ -0,0 +1,7 @@
+#include <lasxintrin.h>
+
+int main(void) {
+  __m256i a, b, c;
+  c = __lasx_xvand_v(a, b);
+  return 0;
+}
```
SixWeining, line 7: How about renaming this file to `SimpleProgramCallLSXBuiltin.c` which is more semantically consistent?

SixWeining: Another alternative is writing this temporary file when running cmake. We can refer the `try_run` usage in llvm-project: `llvm-project/llvm/cmake/modules/FindZ3.cmake`.

The pros is we can keep the `cmake/modules` directory as clean as possible.

leecheechen:
Address @SixWeining's comments and thanks for your feedback.

SixWeining (approved):

PR #82, opened 2024-01-26
[flang]enable execute_command_line tests

yi-wu-arm:
build and pass all three tests on Linux, will test it on Windows later today

### DIFF
--- a/Fortran/gfortran/regression/DisabledFiles.cmake
+++ b/Fortran/gfortran/regression/DisabledFiles.cmake
@@ -844,11 +844,6 @@ file(GLOB UNIMPLEMENTED_FILES CONFIGURE_DEPENDS
   erf_2.F90
   erf_3.F90
 
-  # unimplemented: intrinsic: execute_command_line
-  execute_command_line_1.f90
-  execute_command_line_2.f90
-  execute_command_line_3.f90
-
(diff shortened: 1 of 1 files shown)


tblah (approved):
LGTM: execute command line was added in https://github.com/llvm/llvm-project/pull/74077

yi-wu-arm:
It will compile and run on Windows, but it does not behave the same. This is because Windows has different command, dir <=> ls, timeout <=> sleep.

tarunprabhu (approved):
Tested on Linux on x86 and it looks to be fine. Thanks for enabling these.

tarunprabhu:
> It will compile and run on Windows, but it does not behave the same. This is because Windows has different command, dir <=> ls, timeout <=> sleep.

Just out of curiosity. What do you mean by this?

yi-wu-arm:
> > It will compile and run on Windows, but it does not behave the same. This is because Windows has different command, dir <=> ls, timeout <=> sleep.
> 
> Just out of curiosity. What do you mean by this?

This is because `execute_command_line` calls `std::system`, and on Windows that is cmd. cmd doesn't have `ls`, the Windows equivalent command is `dir`, and for `timeout` for `sleep`. It will compile and run correct based on the standard, however, the printout from the execution won't be the same on Linux.

PR #83, opened 2024-01-29
[flang] Disable some tests whose errors are now warnings

klausler:
Recent changes to generic interface analysis have reduced a previous hard error message to an optional warning.  These tests must be run with -pedantic -Werror to cause compilation to fail.  Until I can figure out how to make that happen, disable those tests so that build bots don't fail.


vzakhari (approved):
I am okay with this `panic` fix :)

clementval (approved):
LGTM

PR #87, opened 2024-02-01
[AArch64][FMV] Add a non-comprehensive test for ACLE FMV

jroelofs:
This covers the ones we currently have enabled via [apple.inc](https://github.com/llvm/llvm-project/blob/main/compiler-rt/lib/builtins/cpu_model/aarch64/fmv/apple.inc#L30-L58) in compiler-rt, as well as a couple that are not.  Each test uses Function Multi Versioning to check for the presence or absence of each given feature, and subsequently exercises that feature if it is presumed to exist.  For features that are not thought to exist by the FMV infra, it forks and attempts to try the feature anyway, expecting the resulting thread to die, thus affirming the inverse.


ilinpv (commented):
Thanks for compiler-rt FMV tests! We are going to add more on runtime detection implementing ``__builtin_cpu_supports`` aarch64 part https://reviews.llvm.org/D153153 on top of recently merged https://github.com/llvm/llvm-project/pull/68919

SingleSource/UnitTests/AArch64/acle-function-multi-versioning.c
```diff
@@ -0,0 +1,259 @@
...
+    asm volatile (
+        "fmov	h0, #0" "\n"
+        : : : "v0"
+    );
+})
+CHECK(ssbs2, {
+    // FIXME: don't know how to test for this
+})
+CHECK(bti, {
+    // FIXME: don't know how to test for this
```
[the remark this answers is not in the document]
jroelofs: Ok, sounds like we can do some inline asm for that on Linux/Android and syscalls on Darwin/Fuchsia.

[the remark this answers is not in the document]
jroelofs: I'll leave a hole for other platforms to implement this in the `#else` case. I tested this on a Mac, but only eyeballed it on Linux.

SingleSource/UnitTests/AArch64/acle-function-multi-versioning.c
```diff
@@ -0,0 +1,288 @@
...
+        "fmov d0, #0" "\n"
+        "fmov d1, #0" "\n"
+        "sha256h q0, q0, v0.4s" "\n"
+        : : : "v0"
+    );
+})
+CHECK(sha1, {
+    asm volatile (
+        "fmov s0, #0" "\n"
+        "sha1h s0, s0" "\n"
```
ilinpv, line 101: A limitation of current implementation encountered here. We don't have support for +sha1 in codegen at the moment ( we need to address that ) and FMV sha1 mapped to '+fp-armv8,+neon' only  https://github.com/llvm/llvm-project/blob/74fb205876be681d2ca9d8bdd23e171e671df5e1/llvm/include/llvm/TargetParser/AArch64TargetParser.h#L266
So in function code assembly check will raise:
```
test-suite/SingleSource/UnitTests/AArch64/acle-function-multi-versioning.c:101:10: error: instruction requires: sha2
  101 |         "sha1h s0, s0" "\n"
      |          ^
<inline asm>:2:1: note: instantiated into assembly here
    2 | sha1h s0, s0
      | ^

```
As workaround implicit encoding can be used:
`.inst 0x5e280800`

SingleSource/UnitTests/AArch64/acle-function-multi-versioning.c
```diff
@@ -0,0 +1,288 @@
...
+    int x;
+    asm volatile (
+        "ldaprb w0, [%0]"
+        : : "r" (&x) : "w0"
+    );
+})
+CHECK(rcpc2, {
+    int x;
+    asm volatile (
+        "ldapurb w0, [%0]"
```
ilinpv, line 130: The same as above, rcpc2 mapped to +rcpc and the asm check will complain on rcpc2 instruction.

ilinpv (commented):
Good to go with Linux part disabled - we'll take it over and address separately, thank you!

SingleSource/UnitTests/AArch64/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+if(CMAKE_C_COMPILER_ID STREQUAL "Clang")
+  if(ARCH STREQUAL "AArch64")
+    llvm_singlesource(PREFIX "aarch64-")
```
ilinpv, line 3: On Linux we also need to specify ``--rtlib=compiler-rt`` to make FMV enabled in driver

jroelofs: Ok. I'll take out the `#if __linux__` for now, and add an Apple check here for now.

PR #90, opened 2024-02-04
Add a benchmark for aarch64's __init_cpu_features_resolver()

jroelofs:


labrinea:
Thanks for writing these tests. Do they need adaptation for other operating systems?

jroelofs:
Yes. There are some subtle differences in the signature of `__init_cpu_features_resolver()`, where on some platforms it takes a hwcap + `__ifunc_arg_t*`, and others it takes no arguments.

labrinea (approved):
Looks good, thanks! I'll write the equivalent linux tests at some point.

PR #91, opened 2024-02-06
[NFC][Fortran/gfortran] Add script to generate static test configuration

tarunprabhu:
This adds the script to generate the static test configuration files from the DejaGNU annotations from the test sources. The script is only intended to be used when the test files are updated from upstream.

### DIFF
--- a/Fortran/gfortran/utils/update-test-config.py
+++ b/Fortran/gfortran/utils/update-test-config.py
@@ -1,10 +1,10 @@
 #!/usr/bin/env python3
 #
 # This script parses the DejaGNU annotations from the files in the gfortran test
-# suite and updates the static test configuration of the tests. This is required
-# to be run whenever the tests are updated with new tests from upstream
+# suite and updates the static test configuration files. This must
+# be run whenever the tests are updated with new tests from upstream
 # gfortran. There are currently several limitations in the way the annotations
-# are parsed and how they are dealt with when determining the static test
+# are parsed and how they are dealt with in the static test
 # configuration. These are described in inline comments. The format of the
 # static test configuration files is also documented inline.
 #
@@ -24,8 +24,11 @@ import typing
 
 # Class representing a single test. The fields of the test should be those that
 # are eventually serialized into the test configuration. The configuration will
-# only contain the test kind, the sources and flags that indicate what DejaGNU
-# annotations are present in the test file.
+# only contain the test kind, the sources and flags as determined directly
+# from the DejaGNU annotations in the corresponding source file(s). Any custom
+# handling of the test, e.g. to run it conditionally on some platform, should
+# not be present, either in this class or in the generated static test
+# configuration file.
 class Test:
     # int, [os.path], [str], [str], bool
     def __init__(
@@ -44,7 +47,7 @@ class Test:
         # The sources needed by the test. This will have at least one element.
         # The first element of the list will be the "main" file. The rest must
         # be in the order in which they should be compiled. The elements will be
-        # the basename's of the file because all dependent files are in the
+        # the basenames of the files because all dependent files are in the
         # same directory, so there is no need to have the full (or relative)
         # path.
         self.sources: list[str] = sources
@@ -71,10 +74,6 @@ class Test:
         # have been captured in the self.enabled_on member of this class.
         self.xfail: bool = expected_fail
 
-        # Whether to use separate compilation for this test (is this even
-        # necessary?)
-        self.separate_compilation = False
-
     def __eq__(self, other):
         if not isinstance(other, Test):
             return NotImplemented
@@ -86,8 +85,6 @@ class Test:
             self.enabled_on == other.enabled_on and \
             self.disabled_on == other.disabled_on
 
-    # The string-ified test is in comma-separated because that's all we need
-    # from this.
     def __str__(self):
         return ';'.join([
             self.kind,
@@ -187,6 +184,15 @@ def qsplit(s: str) -> list[str]:
         s = s[:-1]
     return s.split()
 
+# Drop the leading '{' and trailing '}', if any. This will only drop the
+# braces if both are present. The string will be unconditionally stripped of
+# leading and trailing whitespace.
+def strip_braces(s: str) -> str:
+    s = s.strip()
+    if s.startswith('{') and s.endswith('}'):
+        s = s[1:-1].strip()
+    return s
+
 # Try to match the line with the regex. If the line matches, add the match
 # object to the MOUT list and return True. Otherwise, leave the MOUT list
 # unchanged and return False.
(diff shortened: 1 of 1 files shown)


tblah (approved):
When I ran this myself, `achar_2.f90` was changed from a `run` to a `compile` test. The output complained that it could not open the file. If this cannot be reproduced on your machine then don't let it block the patch.

Otherwise, this looks good. I'm not an expert in python or dejagnu, but the code seems well structured and easy to follow. The `FIXME`s seem reasonable to leave for a later date because this can already generate a test configuration that is an improvement upon what we had before.

Thank you for your work on this!

Fortran/gfortran/utils/update-test-config.py
```diff
@@ -0,0 +1,633 @@
...
+    t: str, enabled_on: list[str], disabled_on: list[str], options: list[str]
+) -> None:
+    t = t.strip()
+
+    # An expression can be wrapped with braces. While this seems to be necessary
+    # for complex expressions, it can be used with simple expressions as well.
+    if t.startswith('{') and t.endswith('}'):
+        t = t[1:-1].strip()
+
+    # A complex expression is one which does not have any logical operators.
```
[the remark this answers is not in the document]
tarunprabhu: It was indeed. Nice catch!

Fortran/gfortran/utils/update-test-config.py
```diff
@@ -0,0 +1,633 @@
...
+# object to the MOUT list and return True. Otherwise, leave the MOUT list
+# unchanged and return False.
+def try_match(regex: re.Pattern, line: str, mout: list[re.Match[str]]) -> bool:
+    m = regex.search(line)
+    if m:
+        mout.append(m)
+        return True
+    return False
+
+# Count the number of elements in the list that match satisfy the predicate.
```
luporl, line 200:
```suggestion
# Count the number of elements in the list that satisfy the predicate.
```

Fortran/gfortran/utils/update-test-config.py
```diff
@@ -0,0 +1,633 @@
...
+def count_if(l, predicate):
+    return sum(1 for e in l if predicate(e))
+
+# Print a message. This is only around to save a bit of typing.
+def printf(fmt: str, *args) -> None:
+    print(fmt.format(*args))
+
+# Print a message in verbose mode.
+def message(fmt: str, *args) -> None:
+    printf(fmt, *args)
```
luporl, line 210: Should this be controlled by some verbose flag?

Fortran/gfortran/utils/update-test-config.py
```diff
@@ -0,0 +1,633 @@
...
+            targets.append(convert_target_regex(platforms[tgt]))
+        # Some "targets" need to be translated to compiler/linker flags.
+        elif tgt in ['fopenmp', 'fopenacc', 'pthread']:
+            options.append('-' + tgt)
+        elif tgt in ['c99_runtime']:
+            options.append('-lc')
+        elif tgt in [
+            'fd_truncate',
+            'fortran_large_int',
+            'fortran_real_10',
```
luporl, line 265: I think `real(10)` is supported on x86 only. But it should be ok to ignore this for now.

Fortran/gfortran/utils/update-test-config.py
```diff
@@ -0,0 +1,633 @@
...
+# list of flags that are passed in.
+def parse_targets_into(
+    t: str, enabled_on: list[str], disabled_on: list[str], options: list[str]
+) -> None:
+    t = t.strip()
+
+    # An expression can be wrapped with braces. While this seems to be necessary
+    # for complex expressions, it can be used with simple expressions as well.
+    if t.startswith('{') and t.endswith('}'):
+        t = t[1:-1].strip()
```
luporl, line 321: This could be moved to a `strip_braces` helper function, since it's used in 3 functions.

luporl (approved):
LGTM. I have only a couple nits and suggestions, but feel free to ignore them.

tarunprabhu:
> > When I ran this myself, `achar_2.f90` was changed from a `run` to a `compile` test. The output complained that it could not open the file. If this cannot be reproduced on your machine then don't let it block the patch.
> 
> Strange. I'll check again on my end.

@tblah, It seems to be working fine for me.

Fortran/gfortran/utils/update-test-config.py
tarunprabhu: Thanks. I didn't know that. I might need some help when I get around to fixing that.

Fortran/gfortran/utils/update-test-config.py
tarunprabhu: This is an obsolete comment. At some point, I had a verbose flag, but I removed it because it seemed unnecessary.

Fortran/gfortran/utils/update-test-config.py
```diff
@@ -0,0 +1,633 @@
...
+        # Whether the test is expected to fail. For run tests, this indicates
+        # the presence of a shouldfail annotation. For all other test kinds,
+        # a dg-error annotation is present somewhere in the file. In the latter
+        # case, the error may only manifest on certain targets, but that should
+        # have been captured in the self.enabled_on member of this class.
+        self.xfail: bool = expected_fail
+
+        # Whether to use separate compilation for this test (is this even
+        # necessary?)
+        self.separate_compilation = False
```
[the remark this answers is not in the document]
tarunprabhu: I think I added this at some point when I wasn't sure exactly what the format of the test configuration files would be. Currently, for multi-file tests, all the Fortran sources are compiled in a single invocation of the compiler. The only time separate compilation is used is if any of the files are not Fortran. The flag was intended to be assert that separate compilation should always be used. In hindsight, I don't think that is necessary, so I have removed it.

Fortran/gfortran/utils/update-test-config.py
```diff
@@ -0,0 +1,633 @@
...
+import re
+import shutil
+import typing
+
+# Class representing a single test. The fields of the test should be those that
+# are eventually serialized into the test configuration. The configuration will
+# only contain the test kind, the sources and flags that indicate what DejaGNU
+# annotations are present in the test file.
+class Test:
+    # int, [os.path], [str], [str], bool
```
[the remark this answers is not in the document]
tarunprabhu: I have rephrased the comment in my latest commit. Please take a look and let me know if it is clearer now. If not, I can reword it again.

tarunprabhu:
@luporl Thanks for the careful review. 

 I have reworded some of the comments in the code. Could you take a look and let me know if they are clearer now?

luporl (approved):
@tarunprabhu Thanks for taking the time for addressing my comments, everything is clearer now.

DavidSpickett:
> When I ran this myself, achar_2.f90 was changed from a run to a compile test. The output complained that it could not open the file. If this cannot be reproduced on your machine then don't let it block the patch.

I hit this too. achar_2.f90 has a strange encoding and Python is maybe detecting it incorrectly. I'm looking into it.

```
WARNING: Could not open file: /home/davspi01/llvm-test-suite/Fortran/gfortran/regression/achar_2.f90 ('charmap' codec can't decode byte 0x81 in position 6415: character maps to <undefined>) Windows-1254
```

DavidSpickett:
https://github.com/llvm/llvm-test-suite/pull/488

PR #93, opened 2024-02-12
[Fortran] Enable tests that failed when using libpgmath on AArch64 Linux

DavidSpickett:
Now that we're not using libpgmath, these tests pass on AArch64.

They may work on Windows now but I don't have a build setup to check yet, and none of the builders run the test suite yet. If/when I can confirm, I'll come back and remove the Windows check too.


DavidSpickett:
Windows opt out was added in https://github.com/llvm/llvm-test-suite/pull/69. @omjavaid what's the status of running the test suite on Windows, can I just use what's upstream or is there work left to do?

antmox (approved):
LGTM

DavidSpickett:
We'll know very quickly what the buildbots think, so I'm going to land this and follow up with Omair later for the Windows side.

PR #94, opened 2024-02-13
[SingleSource/Atomic] Add preliminary tests for atomic builtins.

Logikable:
There exist atomic IR unit tests and libatomic unit tests, but neither can test the atomicity and interoperability of atomic builtins and compiler-rt's atomic library. These tests aim to approximate behaviour encountered in user code.

These tests have caught issues in Clang. See llvm/llvm-project#74349 and llvm/llvm-project#73176 for LLVM changes inspired by these tests.

Note to reviewers: this is a recreation of the accidentally closed https://github.com/llvm/llvm-test-suite/pull/78.

### DIFF
--- a/SingleSource/UnitTests/Atomic/CMakeLists.txt
+++ b/SingleSource/UnitTests/Atomic/CMakeLists.txt
@@ -1,0 +1,9 @@
+# Link the Clang built libatomic.
+execute_process(COMMAND ${CMAKE_C_COMPILER} --print-file-name=libclang_rt.atomic.so
+                OUTPUT_VARIABLE _path_to_libatomic
+                OUTPUT_STRIP_TRAILING_WHITESPACE)
+get_filename_component(_libatomic_dir ${_path_to_libatomic} DIRECTORY)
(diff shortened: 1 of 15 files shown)


SingleSource/UnitTests/Atomic/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+# Link the Clang built libatomic.
+execute_process(COMMAND ${CMAKE_C_COMPILER} --print-file-name=libclang_rt.atomic.so
```
jyknight, line 2: That library isn't built by default, so almost nobody will have a copy. I assume we need to have some conditionals here to skip testing if the library isn't present? 

Potentially we should additionally integration-test against `-latomic` (even if that comes via GCC), since that's what real users actually specify.

jyknight (commented):
Overall looks good to me, modulo the one comment/question.

SingleSource/UnitTests/Atomic/CMakeLists.txt
Logikable: Done. I'm not great with CMake -- what are better ways to do what I have here?

Logikable:
Gentle bump.

PR #95, opened 2024-02-15
[Fortran/gfortran][NFC] Update erroring FFLAGS list

tarunprabhu:
Add more flags that are used in the gfortran test suite but are not supported by flang to the list of blacklisted flags.

-----------------

This should not affect the tests as they are run currently, but will be necessary when the static test configuration are used.

### DIFF
--- a/Fortran/gfortran/CMakeLists.txt
+++ b/Fortran/gfortran/CMakeLists.txt
@@ -62,6 +62,13 @@ set(TEST_SUITE_FORTRAN_ISO_C_HEADER_DIR "" CACHE STRING
 # The following cause errors if they are passed to flang via FFLAGS. This could
 # be because they are currently unsupported and might eventually be supported
 # or because they are GCC-specific and will never be supported.
+#
+# FIXME: The flags here necessarily include the value as well, not just the
+# flag itself. For instance, in the list are -fcheck=all, -fcheck=bits etc. In
+# principle, one only needs the -fcheck. The argument should not matter. The
+# problem is that the annotation parser currently does not parse the flag into
+# `flag-name` and `flag-value`. Once this is fixed, we need only have the
+# flag name here.
 set(FLANG_ERRORING_FFLAGS
   # The "flags" that do not start with a hyphen (-) are actually arguments to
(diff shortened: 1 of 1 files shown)


Fortran/gfortran/CMakeLists.txt
```diff
@@ -63,18 +63,62 @@ set(TEST_SUITE_FORTRAN_ISO_C_HEADER_DIR "" CACHE STRING
...
+  # "-flag=value" syntax). Since we don't actually parse the command line, they
+  # are treated as if they were "flags" .
+  analyzer-max-svalue-depth=0
+  ggc-min-expand=0
+  ggc-min-heapsize=0
+  iv-max-considered-uses=2
+  large-stack-frame=4000
   max-completely-peel-loop-nest-depth=1
+  max-completely-peeled-insns=0
+  max-completely-peeled-insns=200
```
luporl, line 77: Why does `max-completely-peeled-insns` appear twice?
Is `max-completely-peeled-insns` only not supported with values 0 and 200?
If not, shouldn't it (and the other flags) always be ignored, independently of its value?

tarunprabhu: The parser for the DejaGNU attributes in `utils/update-test-config.py` does not currently deal with command line flags correctly. It simply splits the annotation on whitespace and treats each substring as a "flag". It therefore cannot tell the difference between `--flag=val1` and `--flag=val2`. There are other consequences of this. At the start of the list there are entries which do not begin with `-`. This is because they are written in the annotations as `--flag val` and the parser does not know that this is equivalent to `--flag=val`. 

Making the parser a bit smarter about flags is on my to-do list.

luporl: Ok. Then maybe a TODO comment could be added, briefly explaining the current limitations of the parser, and that later this blacklist could be simplified, by removing the flags' values and duplicates.

tarunprabhu: Done.

luporl (approved):
LGTM

PR #96, opened 2024-02-20
utils/compare.py: Fix syntax warning on python 3.12

lukel97:
As of Python 3.12 there's now a syntax warning for invalid escape sequences:

    ./utils/compare.py:195: SyntaxWarning: invalid escape sequence '\g'

This fixes it by using a raw string literal, as recommended here:
https://docs.python.org/3/library/re.html


MatzeB (approved):
LGTM

PR #98, opened 2024-02-21
[AArch64][FMV] Enable tests for ACLE FMV on Linux.

ilinpv:
Enabling Linux part of https://github.com/llvm/llvm-test-suite/pull/87


SingleSource/UnitTests/AArch64/acle-fmv-features.c
```diff
@@ -196,7 +196,14 @@ CHECK(ssbs2, {
 CHECK(bti, {
     // The only test for this requires reading a register that is only
     // accessible to EL1.
-    #if defined(__APPLE__)
+    #ifdef __linux__
```
jroelofs, line 199: Could use something like this as a continuation of the comment above:
```suggestion
    #ifdef __linux__
        // On linux, the kernel intercepts the trap, and emulates it for EL0 processes.
```

SingleSource/UnitTests/AArch64/CMakeLists.txt
```diff
@@ -1,10 +1,9 @@
...
+  if(TARGET_OS STREQUAL "Linux")
+    list(APPEND CFLAGS --rtlib=compiler-rt)
+    list(APPEND LDFLAGS --rtlib=compiler-rt)
+  endif()
   if(ARCH STREQUAL "AArch64")
-    set(Source )
-    # TODO: this test is currently only supported on Darwin platforms:
-    if (TARGET_OS MATCHES "Darwin")
-      list(APPEND Source acle-fmv-features.c)
-    endif()
```
jroelofs, line 7: Does it get added to `Sources` automatically for all platforms now?

ilinpv: I think so looking into https://github.com/llvm/llvm-test-suite/blob/458fed38c2fb9db63b6dd8ced9708e557f3657ff/SingleSource/UnitTests/CMakeLists.txt#L11
But I tested that on aarch64/x86 Linux only

jroelofs: ok, seems fine. If we have trouble with non-posix systems (Fuchsia? Windows?) I guess those can be opted out if the buildbots complain.

jroelofs (approved):

SingleSource/UnitTests/AArch64/CMakeLists.txt
```diff
@@ -1,10 +1,9 @@
 if(CMAKE_C_COMPILER_ID STREQUAL "Clang")
+  if(TARGET_OS STREQUAL "Linux")
+    list(APPEND CFLAGS --rtlib=compiler-rt)
+    list(APPEND LDFLAGS --rtlib=compiler-rt)
```
DavidSpickett, line 4: Guessing this might be an issue if the compiler does include / has not built compiler-rt.

ilinpv: Thank you for report, compiler-rt is required for these tests to work on Linux. Is there a clang configuration without compiler-rt used? Not sure do we need to detect in the tests that compiler-rt is missing and not include them in testing or just make compiler-rt build part of buildbot configuration.

DavidSpickett: I found the rest of the linker output:
```
/usr/bin/ld: cannot find /home/tcwg-buildbot/worker/clang-aarch64-global-isel/stage1.install/lib/clang/19/lib/linux/libclang_rt.builtins-aarch64.a: No such file or directory
```
Which confirms it.

There's no requirement to have compiler-rt present and it seems like this is the first set of tests that need it. A script was recently added to find it on Windows https://github.com/llvm/llvm-test-suite/commit/1c4054cda38233a71767832048a28a9e4c41adaa so I take from that that it has never been needed on Linux before.

So if that script also works for Linux you could call that and skip if we don't find it. @omjavaid would that work on Linux too?

The other option is a "does compiler support --rtlib=compiler-rt" check, in the normal cmake fashion.

(and coverage wise, we have other AArch64 bots that do build compiler-rt that are running the test, so not running it on this one isn't a problem)

omjavaid: Yes I think find_compiler_rt_library in llvm-test-suite/cmake/modules/HandleCompilerRT.cmake should work for linux. The function is derived from llvm-project/cmake/modules/HandleCompilerRT.cmake which works for Linux as well. some minor changes are may be required and can be referenced from to the llvm-project version of HandleCompilerRT.cmake module

PR #103, opened 2024-03-01
[flang] llvm-test-suite/Fortran/gfortran/regression/oldstyle_5.f comp…

klausler:
…iles

This test now compiles and should no longer be expected to fail.


klausler:
This PR will be merged at the same time as the patch to llvm-project that makes this test compile.

psteinfeld (approved):
Looks good.

PR #104, opened 2024-03-05
[flang] Disable pr36006-2 on aarch64.

vzakhari:
It used to fail due to the missing runtime support, but this
was not considered a failure.
It fails due to https://github.com/llvm/llvm-project/issues/84088 now.

https://lab.llvm.org/buildbot/#/builders/184/builds/10735

### DIFF
--- a/Fortran/gfortran/regression/tests.cmake
+++ b/Fortran/gfortran/regression/tests.cmake
@@ -2192,7 +2192,7 @@ compile;pr34163.f90;;-O2 -fno-tree-pre -fpredictive-commoning -fdump-tree-pcom-d
 compile;pr35031.f90;xfail;;;
 compile;pr35849.f90;xfail;;;
 compile;pr36006-1.f90;;;;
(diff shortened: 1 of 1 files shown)


psteinfeld (commented):
Looks good!

vzakhari:
Thank you, Pete! I am not sure what way of disabling the test I should use.  It looks like this one did not work.  I tried #105 and waiting for buildbots to pick it up.

vzakhari:
It looks like #150 fixed the buildbots, so changes in `tests.cmake` does not seem to make any effect now. I also noticed that `tests.cmake` is auto-generated, so I should not have changed it in the first place :)

tarunprabhu:
> so I should not have changed it in the first place :)

Yes :-) Thanks for #105 

The `tests.cmake` files are intended to contain *only* the parsed DejaGNU annotations. Any additional "tweaks" - for instance to disable tests on certain platforms are intended to be done separately. I will add that feature next.

PR #105, opened 2024-03-06
[flang] Disable pr36006-2.

vzakhari:
It used to fail due to the missing runtime support, but this
was not considered a failure.
It fails due to llvm/llvm-project#84088 now.

https://lab.llvm.org/buildbot/#/builders/184/builds/10735

This also reverts #104.


tarunprabhu:
Thanks for this @vzakhari. Apologies for not seeing this sooner.

PR #116, opened 2024-04-18
[FMV] Remove unsupported tests.

labrinea:
In ACLE we are deprecating a bunch of FMV features (see https://github.com/ARM-software/acle/pull/315), so in this patch I am removing the corresponding tests.

### DIFF
--- a/SingleSource/UnitTests/AArch64/acle-fmv-features.c
+++ b/SingleSource/UnitTests/AArch64/acle-fmv-features.c
@@ -95,26 +95,11 @@ CHECK(sha2, {
         : : : "v0"
     );
 })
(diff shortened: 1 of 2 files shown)


labrinea:
The patch [#88965](https://github.com/llvm/llvm-project/pull/88965) broke the build and was reverted with [#89184](https://github.com/llvm/llvm-project/pull/89184), hence I am removing these tests to reland the former.

jroelofs (approved):

labrinea:
Putting on hold for the same reasons explained here https://github.com/llvm/llvm-project/pull/89232#issuecomment-2064303120

PR #117, opened 2024-04-30
[Fortran/gfortran] Disable OpenACC tests that cause non-deterministic segfaults

tarunprabhu:
https://github.com/llvm/llvm-project/issues/90477

These were also observed to cause non-deterministic out of memory errors.


kiranchandramohan (approved):
Thanks Tarun. LGTM.

tblah (approved):
Thanks!

PR #118, opened 2024-05-03
[SPEC] Workaround error with latest clang in 523.xalancbmk_r

lukel97:
xalan-c contains a latent error in a template that isn't instantiated, which after https://github.com/llvm/llvm-project/pull/90152 will now fail to build. We can workaround it by setting -fdelayed-template-parsing. See https://github.com/llvm/llvm-project/pull/90152#issuecomment-2089786180

### DIFF
--- a/External/SPEC/CINT2017rate/523.xalancbmk_r/CMakeLists.txt
+++ b/External/SPEC/CINT2017rate/523.xalancbmk_r/CMakeLists.txt
@@ -10,8 +10,8 @@ set(CMAKE_CXX_STANDARD 14)
 
 include(CheckCXXCompilerFlag)
 check_cxx_compiler_flag(-fdelayed-template-parsing
-                        HAVE_CXX_FLAG_FDELAYED_PARSING)
-if(HAVE_CXX_FLAG_FDELAYED_PARSING)
+                        HAVE_CXX_FLAG_FDELAYED_TEMPLATE_PARSING)
+if(HAVE_CXX_FLAG_FDELAYED_TEMPLATE_PARSING)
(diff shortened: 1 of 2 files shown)


External/SPEC/CINT2017rate/523.xalancbmk_r/CMakeLists.txt
```diff
@@ -8,6 +8,8 @@ speccpu2017_benchmark(RATE)
 
 set(CMAKE_CXX_STANDARD 14)
 
+add_compile_options(-fdelayed-template-parsing)
```
fhahn, line 11: Do we need to check if the flag is available to not break when building with compilers that don't have the flag (looks like it get added to Clang in 2011, so we should be good there, but llvm-test-suite also supports building with other compilers, which may not have the flag)

fhahn (commented):
Thanks for the patch.

Looks like there's the same issue with the version in SPEC2006. Could you also fix that (External/SPEC/CINT2006/483.xalancbmk/CMakeLists.tx)?

External/SPEC/CINT2017rate/523.xalancbmk_r/CMakeLists.txt
lukel97: Good idea, done

lukel97:
> Looks like there's the same issue with the version in SPEC2006.

I've added it to 483.xalancbmk, but I don't have a copy of SPEC2006 to test it out with. Would you be able to give it a quick check that it works?

fhahn (approved):
LGTM, thanks!

PR #122, opened 2024-05-09
[flang] Disable three error tests

klausler:
Some interoperability checks are being relaxed in an upcoming change to the compiler, and an error has been downgraded to a portability warning.  This will cause these three tests to "fail" due to lack of a fatal error, so I'm disabling them now to prevent build bot failures.

If I knew how to run these tests with "-pedantic -Werror", I would do that instead.


tarunprabhu (approved):
I intend to add some support to allow a custom set of flags to be added when building a test since this is an issue with some other tests as well.  

Thanks for this.

klausler:
Thanks for the very quick review!

PR #126, opened 2024-05-27
[AIX][z/OS] Disable strip before hashing

daltenty:
Similar to other platforms already opted out, these platforms strip utility does support the extra options used here.


abhina-sree (approved):
LGTM

PR #128, opened 2024-06-03
Adjust tests for llvm commit 4b163e343cfa

sks75220:
That commit changed the mangling of some existing template parameters. This commit adjusts the tests to the new mangling, but keeps old tests under -fclang-abi-compat=17.


wjristow (approved):
LGTM

PR #129, opened 2024-06-04
[sollve_vv] Autodetected native GPU

Meinersbur:
Using the CMake option `-DTEST_SUITE_SYSTEM_GPU=native` or (`-DTEST_SUITE_SYSTEM_GPU=auto`), automatically pre-selects the tests that are expected to work based on the system's GPU.

Also fixes a lot of other issues:
 * `SYSTEM_GPU` violates naming conventions
 * `SYSTEM_GPU` is not a cached variable, i.e. after re-running CMake (e.g. after changing a `CMakeLists.txt`), its value is forgotten
 * Remove `intel` and `regression` options. There are no lists for these
 * Remove possibility to specify multiple pre-select lists. The union of those would run, causing them to fail if run on the other vendor's GPU.
 * Remove duplicate `test_metadirective_arch_is_nvidia.c` in preselect lists
 * Update documentation
 * Add messages whenever SOLLVE V&V is skipped
 * Do not dump test directories into the builddir root
 * Replace `TEST_SUITE_FORCE_ALL` with `-DTEST_SUITE_SYSTEM_GPU=all`
 * Remove uses `remove_definitions`

The variable `SYSTEM_GPU` is also used in External/HeCBench and External/smoke. These are based on sollve_vv's previous `CMakeLists.txt` and are not affected.


External/OpenMPOffloading.cmake
```diff
@@ -0,0 +1,73 @@
...
+  return ()
+endif ()
+
+add_library(OpenMP_Offloading INTERFACE IMPORTED)
+foreach (_lang IN ITEMS C CXX Fortran)
+  if (lang STREQUAL "Fortran" AND NOT TEST_SUITE_FORTRAN)
+    continue ()
+  endif ()
+  set(TEST_SUITE_OFFLOADING_${_lang}_FLAGS   "--offload-arch=native" CACHE STRING "Compiler arguments for OpenMP offloading in ${_lang}")
+  set(TEST_SUITE_OFFLOADING_${_lang}_LDFLAGS "--offload-arch=native" CACHE STRING "Linker arguments for OpenMP offloading in ${_lang}")
```
jplehr, line 65: I find it a surprising / confusing that here the `offload-arch` is specified as `native`, given that we use the same identifier in all the CMake before with AMD/NVidia checks etc.

jplehr (commented):
After I read the README my comment about being confusing is less relevant (maybe still worth a comment in the CMake file?).

I'm not a CMake guru by any means, but the changes seem reasonable to me.

External/sollve_vv/sollvevv_nvidia_tests.cmake
```diff
@@ -1,9 +1,5 @@
-# SOLLVE OpenMP Offloading Validation & Verification Suite
-# https://crpl.cis.udel.edu/ompvvsollve/
-# Maintained By: Aaron Jarmusch <Jarmusch@udel.edu>
-
-
-set(NVIDIA
+message(STATUS "Using SOLLVE V&V tests known to work with Clang/Flang on AMD GPUs")
```
ajarmusch, line 1: this says AMD GPUs, where it should be NVIDIA

Meinersbur:
> After I read the README my comment about being confusing is less relevant (maybe still worth a comment in the CMake file?).

What were you confused about? In both uses `native` detects what GPU the system has installed. Could you propose such a comment?

External/OpenMPOffloading.cmake
```diff
@@ -0,0 +1,78 @@
...
+  if (TARGET OpenMP::OpenMP_${_lang})
+    target_link_libraries(OpenMP_Offloading INTERFACE OpenMP::OpenMP_${_lang})
+  endif ()
+  separate_arguments(_flags NATIVE_COMMAND ${TEST_SUITE_OFFLOADING_${_lang}_FLAGS})
+  separate_arguments(_ldflags NATIVE_COMMAND ${TEST_SUITE_OFFLOADING_${_lang}_LDFLAGS})
```
ajarmusch, line 75: When testing, these lines 74-75 give an error 

`CMake Error at External/OpenMPOffloading.cmake:74 (separate_arguments):
  separate_arguments given unexpected argument(s)
  External/sollve_vv/CMakeLists.txt:16 (include)
`

This could be that this is a new feature in CMAKE Version 3.9. And the root CMakeList.txt only requires `cmake_minimum_required(VERSION 3.13.4)`. 

Either (1) put `cmake_minimum_required(VERSION 3.9)` at the top of this file. (2) or maybe change the separate_arguments mode to `UNIX_COMMAND` or `WINDOWS_COMMAND`. With a checker for which mode to use. 

Personally I don't mind either or if you have a better idea, tho we might want to specify it somewhere

ajarmusch: the documentation [here ](https://cmake.org/cmake/help/latest/command/separate_arguments.html) has 3.9 but only 3.29 was released

Meinersbur: 3.13.4 is later than 3.9. I think that the last argument must be put into quotes, as the command expect exactly 3 arguments (not e.g. 4 when `TEST_SUITE_OFFLOADING_C_FLAGS` is `--offload-arch=native;--cuda-path=/usr/local`). I will experiment.

ajarmusch: ah - I had to remove the semi-colon in the flag `TEST_SUITE_OFFLOADING_C_FLAGS="-fopenmp --offload-arch=native"` to get passed the error

jplehr:
> > After I read the README my comment about being confusing is less relevant (maybe still worth a comment in the CMake file?).
> 
> What were you confused about? In both uses `native` detects what GPU the system has installed. Could you propose such a comment?

My confusion comes from the `--offload-arch=native` flag, which may generate code for multiple architectures in case they are present, IIRC.
OTOH I think the `native` CMake variable is used to select one of AMD or NVidia. Did I misread this?

How would that handle systems that have both AMD and NVidia GPUs installed?

Meinersbur:
> How would that handle systems that have both AMD and NVidia GPUs installed?

One of them gets prioritized, in this case it's AMD. 

I don't think multiple GPUs from different vendors is a common setup that would be useful to handle; we don't even have a mechanism in the test-suite to select which one to run on. If there is, the user will have to explicitly select one with `--offload-arch` and `-DTEST_SUITE_SYSTEM_GPU`, with different build dirs for each GPU.

ajarmusch:
> > How would that handle systems that have both AMD and NVidia GPUs installed?
> 
> 
> 
> One of them gets prioritized, in this case it's AMD. 
> 
> 
> 
> I don't think multiple GPUs from different vendors is a common setup that would be useful to handle; we don't even have a mechanism in the test-suite to select which one to run on. If there is, the user will have to explicitly select one with `--offload-arch` and `-DTEST_SUITE_SYSTEM_GPU`, with different build dirs for each GPU.

Currently, we are running the test-suite in a pipeline on a system with AMD and NVIDIA GPUs. Though, we just run it twice while specifically setting the GPU we want to run. 

Since this is uncommon, giving the user the option to select which hardware they want is reasonable enough.

Meinersbur:
> Currently, we are running the test-suite in a pipeline on a system with AMD and NVIDIA GPUs. Though, we just run it twice while specifically setting the GPU we want to run.

Do you currently set `-DSYSTEM_GPU=amd,nvidia`? Do you invoke `lit`/`make check` twice with a different `OMP_DEFAULT_DEVICE` to make tests run on different devices? Even then, before the patch `-DSYSTEM_GPU=amd,nvidia` would execute the union of both greenlists, so tests will fail on the other. This doesn't seem ideal.

ajarmusch:
> Do you currently set `-DSYSTEM_GPU=amd;nvidia`? Do you invoke `lit`/`make check` twice with a different `OMP_DEFAULT_DEVICE` to make tests run on different devices? Even then, before the patch `-DSYSTEM_GPU=amd;nvidia` would execute the union of both greenlists, so tests will fail on the other. This doesn't seem ideal.

so first we build LLVM twice using `LIBOMPTARGET_PLUGINS_TO_BUILD` to build the NVIDIA or AMD plugin into different dirs. we also build the test-suite twice in different dirs, once with `-DSYSTEM_GPU=amd` and once with `-DSYSTEM_GPU=nvidia`. our thought was to have individual jobs for each GPU

jplehr:
I'm fine with the user being responsible for selecting the right device / build config and combination.

Meinersbur:
> so first we build LLVM twice using `LIBOMPTARGET_PLUGINS_TO_BUILD` to build the NVIDIA or AMD plugin into different dirs. we also build the test-suite twice in different dirs, once with `-DSYSTEM_GPU=amd` and once with `-DSYSTEM_GPU=nvidia`. our thought was to have individual jobs for each GPU

So you don't need `-DSYSTEM_GPU=amd;nvidia` either then? Typically they both are different machines. Systems with dedicated AMD and Nvidia GPUs are not exactly common outside of testing.

ajarmusch:
> So you don't need `-DSYSTEM_GPU=amd;nvidia` either then? Typically they both are different machines. Systems with dedicated AMD and Nvidia GPUs are not exactly common outside of testing.

no, you are right. I don't use both `amd;nvidia` in the same build.

Meinersbur:
ping

ajarmusch (approved):
I am fine with all the changes made

PR #130, opened 2024-06-04
musl compatibility

v01dXYZ:
musl has some discrepancies with glibc that causes some issues at compilation/run:

* headers that don't provide the same set of symbols (`time.h`)
* usage of routines that depends on behaviour left to the implementation by the spec (`random`). At the end, the program generates very different outputs.

Feel free to suggest the right tags for the commits (`[MultiSource]`, `[CLAMR]`, ...).

### DIFF
--- a/MultiSource/Applications/obsequi/Obsequi.reference_output
+++ b/MultiSource/Applications/obsequi/Obsequi.reference_output
@@ -136,10 +136,10 @@ alpha 0, beta 5000.
 cutoffs depth 1: (56) 0 - 0 0 0 0 0 >0.
 The value is -5 at a depth of 1.
 Nodes: 56.
-Move (2,1), value 5000: 2,023,301.
+Move (2,1), value 5000: 2,023,432.
 alpha -5000, beta 5000.
 Winner found: 5000.
-1356 2450 1316053 104829.
+1356 2449 1316027 104839.
 
 cutoffs depth 1: (1) 0 - 0 0 0 0 0 >0.
 cutoffs depth 2: (52) 52 - 52 0 0 0 0 >0.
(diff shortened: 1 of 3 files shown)


v01dXYZ:
I found a discrepancy related to the way buffered IO is implemented by libc.

``` c
#include <stdio.h>
#include <unistd.h>

#define WRITE(fd, str) \
  do { \
  char s[] = str;                               \
  write(fd, s, sizeof(s)); \
  } while (0)

int main(int argc, char **argv)
{
  if(argc < 2) {
    fprintf(stdout, "stdout\n");
    fprintf(stderr, "stderr\n");
  } else {
    WRITE(STDOUT_FILENO, "stdout\n");
    WRITE(STDERR_FILENO, "stderr\n");
  }

  return 0;
}
```

When you use the following command line, with glibc we have

``` shell
$ ./buff_io_discrepancy # no argument -> buffered IO
stderr
stdout
$ ./buff_io_discrepancy direct # argument -> direct IO
stdout
stderr
```

musl outputs the same (in this very case at least, I don't know if it holds always true).

androm3da (approved):

androm3da:
These changes LGTM but is there someone who should approve them?

DavidSpickett (changes requested):
1. This PR has been dormant for a long time, so I want to be careful handling it.
2. For some imported test suites our policy is not to modify the source of the test but to disable it, I don't know how that policy would change with an application. Perhaps we would ideally fix it upstream, if the application is imported from elsewhere.

This is something we should document if we haven't already. @lenary has dealt with this for a test suite before, did we document that decision anywhere?

At least we can make that policy clear, even if this PR does not move forward.

DavidSpickett:
Just to be clear: being compatible with musl seems like a fine goal if there is a need for it.

lenary:
Yeah, documenting "what can be changed" vs "what cannot" has not really been done, and I think some changes have crept in to what is effectively vendored code.

I can maybe write a quick note for the docs this afternoon, if that would help.

androm3da:
> Perhaps we would ideally fix it upstream, if the application is imported from elsewhere.

That makes the most sense.

lenary:
I opened https://github.com/llvm/llvm-project/pull/128937 which I hope adds some clarity about our intentions originally.

nikic:
I believe we only have a hard policy against "don't change the tests" for the GCC torture test suite in particular. We do regularly update other tests to keep them working with new clang versions.

PR #134, opened 2024-06-12
Make code conform with Fortran standard

Leporacanthicus:
The original test "writes" to the argument a of s1, which is a constant value. This isn't valid in Fortran, and will fail if constant values are placed in read-only memory.

The original bug-report wasn't about using a constant argument to a function, but simply that the assign to doesn't work for some cases.

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=22290

This patch fixes the test-code so that it is compliant with Fortran standard, by introducing a variable to assign into.

### DIFF
--- a/Fortran/gfortran/regression/assign_5.f90
+++ b/Fortran/gfortran/regression/assign_5.f90
@@ -10,6 +10,7 @@ go to a
 777 continue
 end
 program test
-call s1 (1)
+integer x
+call s1 (x)
 end
 


tblah (approved):
Agreed. Writing to a literal constant is non-conforming.

tarunprabhu:
For both the GCC and gfortran test suites, we do not make any changes to the tests themselves. This originally was borne out of licensing concerns. I am not particularly keen on changing this stance. 

We have some end to end tests in `Fortran/UnitTests`. We could add a test like this there. `Fortran/UnitTests/hello` is an example.

For this particular test, we should just disable it.

Leporacanthicus:
> For both the GCC and gfortran test suites, we do not make any changes to the tests themselves. This originally was borne out of licensing concerns. I am not particularly keen on changing this stance.
> 
> We have some end to end tests in `Fortran/UnitTests`. We could add a test like this there. `Fortran/UnitTests/hello` is an example.
> 
> For this particular test, we should just disable it.

Ok, I'll add it to the "disable test" list and add something similar to the unit tests.

Leporacanthicus:
Closing, new PR #135 to replace this one.

PR #135, opened 2024-06-12
Disable non-conforming test

Leporacanthicus:
Replace it with our own similar test.

The gfortran/regression/assign_5.f90 writes to a constant value which would cause that test to fail if constants are read-only.

A replacement test assign-goto.f90 is introduced to check that this an assigned goto statement compiles and runs without issue.

### DIFF
--- a/Fortran/gfortran/regression/DisabledFiles.cmake
+++ b/Fortran/gfortran/regression/DisabledFiles.cmake
@@ -1534,6 +1534,10 @@ file(GLOB FAILING_FILES CONFIGURE_DEPENDS
   directive_unroll_5.f90
   # Tests "!GCC$ attributes weak :: x"
   weak-3.f90
+  # Test is not conformant as it writes to a constant argument
+  # Similar test, that is conformant, added to hello.f90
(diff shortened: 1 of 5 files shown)


Leporacanthicus:
This replaces #134 that was "the wrong solution".

Fortran/gfortran/regression/DisabledFiles.cmake
```diff
@@ -1534,6 +1534,10 @@ file(GLOB FAILING_FILES CONFIGURE_DEPENDS
   directive_unroll_5.f90
   # Tests "!GCC$ attributes weak :: x"
   weak-3.f90
+  # Test is not conformant as it writes to a constant argument
+  # Similar test, that is conformant, added to hello.f90
```
tblah, line 1538:
```suggestion
  # Similar test, that is conformant, added to UnitTests/assign-goto
```

tarunprabhu (approved):
Other than the correction to the comment that Tom has identified, LGTM. Tested on Linux on x86-64. 

Thanks for the second PR, Mats.

PR #136, opened 2024-06-21
Allow passing additional flags to the remote execution client

azharudd:
`TEST_SUITE_REMOTE_CLIENT` can be used to pass additional flags to the remote execution client, such as specifying an alternative SSH configuration file.
For example: `TEST_SUITE_REMOTE_CLIENT="ssh -F path/to/config/file"`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -107,7 +107,9 @@ mark_as_advanced(TEST_SUITE_REMOTE_CLIENT)
 if(TEST_SUITE_REMOTE_HOST)
   add_custom_target(rsync
     COMMAND ${PROJECT_SOURCE_DIR}/utils/rsync.sh
-                ${TEST_SUITE_REMOTE_HOST} ${PROJECT_BINARY_DIR}
+                ${TEST_SUITE_REMOTE_CLIENT}
+                ${TEST_SUITE_REMOTE_HOST}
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -107,7 +107,9 @@ mark_as_advanced(TEST_SUITE_REMOTE_CLIENT)
 if(TEST_SUITE_REMOTE_HOST)
   add_custom_target(rsync
     COMMAND ${PROJECT_SOURCE_DIR}/utils/rsync.sh
-                ${TEST_SUITE_REMOTE_HOST} ${PROJECT_BINARY_DIR}
+                ${TEST_SUITE_REMOTE_CLIENT}
```
fhahn, line 110: Do we need to add a default for this somewhere?

azharudd: It is an already existing variable. It is declared at https://github.com/llvm/llvm-test-suite/blob/main/CMakeLists.txt#L103 and defaults to `ssh`.

fhahn: Sounds good, thanks!

fhahn (approved):
LGTM, thanks!

azharudd:
Since this mainly pertains to rsync, I've updated the commit message and added a `[rsync]` tag to it.

PR #137, opened 2024-06-26
update Fortran tests for execute_command_line

yi-wu-arm:
`execute_command_line` was updated in https://github.com/llvm/llvm-project/pull/93023 with more `cmdstat` and richer `cmdmsg`, so existing test will fail.


Fortran/gfortran/regression/DisabledFiles.cmake
```diff
@@ -331,6 +331,11 @@ file(GLOB UNIMPLEMENTED_FILES CONFIGURE_DEPENDS
   erf_2.F90
   erf_3.F90
 
+  # Test is not conformant as it expects different value of cmdstat and cmdmsg
+  # Similar test added: UnitTests/assign-goto
```
tblah, line 335:
```suggestion
  # Similar test added: UnitTests/execute_command_line
```

PR #138, opened 2024-06-28
[flang][tests] Enable llvm-test-suite/Fortran/gfortran/regression/boz…

klausler:
…_8.f90

This test will no longer fail when https://github.com/llvm/llvm-project/pull/96672 is merged into llvm-project/main.


Leporacanthicus (commented):
Looks good to me. I expect Tarun also should approve.

PR #140, opened 2024-07-08
[Fortran] enable gomp tests which now pass

tblah:
These were found through brute force. I haven't checked which PRs fixed them.


tarunprabhu (approved):
Works on Linux on x86-64. 

Thanks for brute-forcing your way through these :-)

PR #142, opened 2024-07-19
[llvm-test-suite][tools] Fix not utility to avoid std::system

tarunprabhu:
Avoid using std::system on Unix systems. Some systems such as Ubuntu return a different error code from calls to std::system. Use posix_spawn to avoid having to special case these. std::system is now only used on Windows.

@jeanPerier identified this as an issue [here](https://github.com/llvm/llvm-test-suite/pull/102#issuecomment-1980674221)
@jroelofs, could you check that this doesn't cause any issues on iOS.

### DIFF
--- a/tools/not.cpp
+++ b/tools/not.cpp
@@ -24,9 +24,12 @@
 #include <windows.h>
 #endif
 
-#ifdef __APPLE__
+#if defined(__unix__) || defined(__APPLE__)
 #include <spawn.h>
 #include <sys/wait.h>
+#endif
+
+#ifdef __APPLE__
 #include <TargetConditionals.h>
 #endif
 
@@ -56,20 +59,23 @@ int main(int argc, char* const* argv) {
     return 1;
 
   int result;
-#if !defined(TARGET_OS_IPHONE)
+#ifdef _WIN32
   std::stringstream ss;
   ss << argv[0];
   for (int i = 1; i < argc; ++i)
     ss << " " << argv[i];
   std::string cmd = ss.str();
   result = std::system(cmd.c_str());
-#else
+#elif defined(__unix__) || defined(__APPLE__)
   pid_t pid;
   if (posix_spawn(&pid, argv[0], NULL, NULL, argv, NULL))
     return EXIT_FAILURE;
   if (waitpid(pid, &result, WUNTRACED | WCONTINUED) == -1)
     return EXIT_FAILURE;
+#else
+  #error "Unsupported system"
 #endif
+
   int retcode = 0;
   int signal = 0;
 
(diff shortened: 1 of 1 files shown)


tools/not.cpp
```diff
@@ -24,9 +24,12 @@
...
 #endif
 
-#ifdef __APPLE__
+#if defined(__unix__) || defined(__APPLE__)
 #include <spawn.h>
 #include <sys/wait.h>
+#endif
+
+#ifdef __APPLE__
 #include <TargetConditionals.h>
```
jroelofs, line 33: `TargetConditionals.h` is only needed for `TARGET_OS_IPHONE`, so I think we can drop this.

jroelofs (approved):

tools/not.cpp
tarunprabhu: Should the `#ifdef` be changed to `TARGET_OS_IPHONE` then?

jeanPerier (approved):
Thank you @tarunprabhu. I am no expert in those if-def, but this patch ran OK on my Ubuntu machine.

tools/not.cpp
```diff
@@ -24,9 +24,12 @@
...
 #endif
```
jroelofs, line 34: I mean:

```suggestion

```

tools/not.cpp
```diff
@@ -56,20 +59,23 @@ int main(int argc, char* const* argv) {
...
   if (posix_spawn(&pid, argv[0], NULL, NULL, argv, NULL))
     return EXIT_FAILURE;
   if (waitpid(pid, &result, WUNTRACED | WCONTINUED) == -1)
     return EXIT_FAILURE;
+#else
+  #error "Unsupported system"
 #endif
+
   int retcode = 0;
   int signal = 0;
```
jroelofs, line 80: I think this should be:

```
  int result;
#if defined(__unix__) || defined(__APPLE__)
  pid_t pid;
  if (posix_spawn(&pid, argv[0], NULL, NULL, argv, NULL))
    return EXIT_FAILURE;
  if (waitpid(pid, &result, WUNTRACED | WCONTINUED) == -1)
    return EXIT_FAILURE;
#else
  std::stringstream ss;
  ss << argv[0];
  for (int i = 1; i < argc; ++i)
    ss << " " << argv[i];
  std::string cmd = ss.str();
  result = std::system(cmd.c_str());
#endif

  int retcode = 0;
  int signal = 0;
```

As I am not sure that `defined(_WIN32) || defined(__unix__) || defined(__APPLE__)` covers everything that _was_ supported prior to this patch.

tarunprabhu: Nice catch. Thanks for taking a close look.

tools/not.cpp
tarunprabhu: Done

jroelofs (approved):

PR #109, opened 2024-03-19
[fortran] Update gfortran disabled test list now that CLASS is lowered

jeanPerier:
All the TODOs and bugs related to CLASS have been implemented/fixed. Enable newly passing tests and move all other failing tests under existing or new categories. None of these failures are bugs, most are different TODOs, unimplemented runtime checks, or cases where flang will behave differently than gfortran on purpose.

PR #143, opened 2024-07-22
[Fortran/gfortran] Enable tests that now pass

tarunprabhu:
This is likely to be dependent on [another PR](https://github.com/llvm/llvm-test-suite/pull/142). 

@jeanPerier attempted to enable several of these tests [once before](https://github.com/llvm/llvm-test-suite/pull/109), but there were issues with the `not` utility in the test suite that prevented the tests from being enabled then. The [first PR linked above](https://github.com/llvm/llvm-test-suite/pull/142) should have fixed those, so once that lands, these tests should also work.

### DIFF
--- a/Fortran/gfortran/regression/DisabledFiles.cmake
+++ b/Fortran/gfortran/regression/DisabledFiles.cmake
@@ -1727,4 +1727,15 @@ file(GLOB FAILING_FILES CONFIGURE_DEPENDS
   # Require -fcheck=bounds or similar.
   bounds_check_17.f90
   pr48958.f90
+
+  # These files require the __truncsfbf2 intrinsic that is not available in
+  # before GCC 13. Alternatively, it requires compiler-rt to be built and a
+  # command line option provided to instruct the compiler to use it. Currently,
+  # we do not support either a version check on GCC or require that compiler-rt
+  # be built, so these are disabled. See:
+  #
+  # https://github.com/llvm/llvm-test-suite/pull/143#discussion_r1689462248
+  #
+  fmt_en.f90
+  fmt_en_rd.f90
 )
--- a/Fortran/gfortran/regression/DisabledFiles.cmake
+++ b/Fortran/gfortran/regression/DisabledFiles.cmake
@@ -1738,4 +1738,7 @@ file(GLOB FAILING_FILES CONFIGURE_DEPENDS
   #
   fmt_en.f90
   fmt_en_rd.f90
+  fmt_en_rn.f90
+  fmt_en_ru.f90
+  fmt_en_rz.f90
 )
--- a/Fortran/gfortran/regression/DisabledFiles.cmake
+++ b/Fortran/gfortran/regression/DisabledFiles.cmake
@@ -466,10 +466,6 @@ file(GLOB SKIPPED_FILES CONFIGURE_DEPENDS
   winapi.f90 # needs -lkernel32 and target *-*-cygwin*
   widechar_11.f90 # No ASSIGNMENT matches TYPE(c_ptr) and TYPE(__builtin_c_ptr)
 
-  # error: 'fir.convert' op invalid type conversion
-  c_char_tests_4.f90
-  c_char_tests_5.f90
-
   # error: not a constant derived type expression
   coarray_42.f90
   init_flag_10.f90
@@ -888,6 +884,8 @@ file(GLOB FAILING_FILES CONFIGURE_DEPENDS
   bounds_check_strlen_4.f90
   bounds_check_strlen_5.f90
   bounds_check_strlen_7.f90
+  c_char_tests_4.f90
+  c_char_tests_5.f90
   char_bounds_check_fail_1.f90
   char_pointer_assign_4.f90
   char_pointer_assign_5.f90
--- a/Fortran/gfortran/regression/DisabledFiles.cmake
+++ b/Fortran/gfortran/regression/DisabledFiles.cmake
@@ -130,7 +130,6 @@ file(GLOB UNIMPLEMENTED_FILES CONFIGURE_DEPENDS
   ISO_Fortran_binding_1.f90
   ISO_Fortran_binding_13.f90
   ISO_Fortran_binding_3.f90
-  ISO_Fortran_binding_6.f90
   PR100029.f90
   PR100097.f90
   PR100098.f90
@@ -179,9 +178,7 @@ file(GLOB UNIMPLEMENTED_FILES CONFIGURE_DEPENDS
   recursive_check_6.f03
 
   # unimplemented: procedure components
-  pr82253.f90
   proc_ptr_24.f90
-  proc_ptr_comp_14.f90
   structure_constructor_11.f90
 
   # unimplemented: procedure pointer results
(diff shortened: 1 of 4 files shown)


Fortran/gfortran/regression/DisabledFiles.cmake
```diff
@@ -524,19 +463,11 @@ file(GLOB SKIPPED_FILES CONFIGURE_DEPENDS
...
-  ISO_Fortran_binding_10.f90 # test reports failure at runtime
-  ISO_Fortran_binding_18.f90 # runtime abort
-  ISO_Fortran_binding_5.f90 # test reports failure at runtime
-  ISO_Fortran_binding_7.f90 # test reports failure at runtime
   winapi.f90 # needs -lkernel32 and target *-*-cygwin*
   widechar_11.f90 # No ASSIGNMENT matches TYPE(c_ptr) and TYPE(__builtin_c_ptr)
 
   # error: 'fir.convert' op invalid type conversion
   c_char_tests_4.f90
-  c_char_tests_5.f90
```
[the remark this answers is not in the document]
tarunprabhu: Done

Fortran/gfortran/regression/DisabledFiles.cmake
```diff
@@ -468,6 +468,7 @@ file(GLOB SKIPPED_FILES CONFIGURE_DEPENDS
 
   # error: 'fir.convert' op invalid type conversion
   c_char_tests_4.f90
+  c_char_tests_5.f90
```
tblah, line 471: nit: this should be moved to the FAILING_FILES section now that it does compile but fails at runtime.

tblah (approved):
Thanks for the update. All passes for me now

Fortran/gfortran/regression/DisabledFiles.cmake
```diff
@@ -661,11 +588,6 @@ file(GLOB SKIPPED_FILES CONFIGURE_DEPENDS
   cray_pointers_7.f90
   dec_math.f90
   dec_math_5.f90
-  fmt_en.f90
-  fmt_en_rd.f90
```
jeanPerier, line 665: I am getting link errors here:
```
/usr/bin/ld: CMakeFiles/gfortran-regression-execute-regression__fmt_en_rd_f90.dir/fmt_en_rd.f90.o: in function `_QFPcheckfmt':
FIRModule:(.text+0x3757): undefined reference to `__truncsfbf2'
/usr/bin/ld: CMakeFiles/gfortran-regression-execute-regression__fmt_en_f90.dir/fmt_en.f90.o: in function `_QFPcheckfmt':
FIRModule:(.text+0x2797): undefined reference to `__truncsfbf2'
```

I am guessing I am likely not installing clang_rt/compiler-rt correctly, there is nothing special to do with regards to compiler-rt when building the test suite, right?

tarunprabhu: > there is nothing special to do with regards to compiler-rt when building the test suite, right?

No, you shouldn't have to do anything special with compiler-rt. I don't build compiler-rt explicitly, so I am not sure why I don't see the same issue. I'll take a look at the binaries on my end too.

Fortran/gfortran/regression/DisabledFiles.cmake
tarunprabhu: Done

Fortran/gfortran/regression/DisabledFiles.cmake
tarunprabhu: I managed to get the same error when `--gcc-install-dir` was not passed and `flang` ended up picking the wrong GCC libraries. Not sure if that is the case with you, but might be worth looking into. 

Let me know how you'd like to proceed.

jeanPerier: Thanks for checking.
> I managed to get the same error when --gcc-install-dir was not passed

Passed to what? When building flang? I tried passing it to flang-new, and it indeed picks the intended gcc version I am using on that machine (9.3.0), but I still get the undefined symbol, so I assume it may not be part of the my gcc lib.

I greped in my 9.3.0 libgcc.a, and I have some hit, but not  `__truncsfbf2`:
````
0000000000000000 T __trunctfsf2
0000000000000000 T __trunctfdf2
0000000000000000 T __trunctfxf2
````

I was able to find  `__truncsfbf2` in libgcc.a 13 and 14, but not 9, 10, 11 and 12.

These symbols are also defined in the compiler-rt builtin library, hence I was wondering if that was how you were getting them. Using `-v` with flang-new you would see libclang_rt.builtins.a or libclang_rt.builtins-x86_64.a be linked , which happens when one use the `--rtlib=compiler-rt` flag with flang-new, but I was wondering if somehow your build of flang-new linked that by default, hence my initial question about compiler-rt. But it seems you are getting them in your gcc environment from what you are saying.

> Let me know how you'd like to proceed.

It appears this change will force the llvm-test-suite to be run with flang built with gcc 13 or higher, which is a significant limitation. I would advise for some cmake option to enable/disable tests requiring such float16 recent builtin symbols.

Comment (tarunprabhu) [reply Fortran/gfortran/regression/DisabledFiles.cmake:665]: > > I managed to get the same error when --gcc-install-dir was not passed
>>
> Passed to what? When building flang?

Ah, sorry, should have been clearer about this. I meant passing `--gcc-install-dir` to `flang-new`. I use GCC 14, which is why I was getting them. If I don't pass `--gcc-install-dir`, I get GCC 8.x, which is also consistent with what you are seeing. 

>  I would advise for some cmake option to enable/disable tests requiring such float16 recent builtin symbols.

I will disable these tests for now until I can figure out some way to selectively enable these depending on the availability of a sufficiently modern GCC or compiler-rt. Thanks for looking into this more closely.

Comment (tarunprabhu) [reply Fortran/gfortran/regression/DisabledFiles.cmake:665]: Disabled the following files

*  fmt_en.f90
*  fmt_en_rd.f90

These were the only ones that were failing, correct?

Comment (jeanPerier) [reply Fortran/gfortran/regression/DisabledFiles.cmake:665]: Thanks for the update. I am hitting the issue on all fmt_en_... tests:
```
  fmt_en.f90
  fmt_en_rd.f90
  fmt_en_rn.f90
  fmt_en_ru.f90
  fmt_en_rz.f90
```

After that, everything passes for me when I apply #142. Thanks!

tarunprabhu: >  I am hitting the issue on all fmt_en_... tests

Updated the list. 

> After that, everything passes for me when I apply https://github.com/llvm/llvm-test-suite/pull/142.

You anticipated my next request. Thanks! 🥇

tarunprabhu:
@tblah, @AlexisPerry,

#142 was just merged. Do you mind checking this again now? Perhaps I'm being a bit overcautious. Thanks!

tblah:
Still works for me

antmox:
Hi @tarunprabhu, @tblah, @jeanPerier,
It looks like some of our bots are broken since this commit:
  clang-aarch64-full-2stage https://lab.llvm.org/buildbot/#/builders/84/builds/480
  clang-aarch64-sve-vla-2stage https://lab.llvm.org/buildbot/#/builders/41/builds/903
  clang-aarch64-sve-vls-2stage https://lab.llvm.org/buildbot/#/builders/4/builds/937
  clang-aarch64-sve-vla https://lab.llvm.org/buildbot/#/builders/17/builds/1349
  clang-aarch64-sve-vls https://lab.llvm.org/buildbot/#/builders/143/builds/917
I still dont understand why all these NOEXE errors, but it seems that at least internal_dummy_2.f08 cannot be re-enabled.
Could you please look at this ?

tarunprabhu:
I'm taking a look now.

tarunprabhu:
@antmox

The `NOEXE` errors don't seem to be coming from the fortran tests, so I can't really speak to those. I could not replicate the error on the AArch64 machine to which I have access. 

Based on an exchange with @tblah and @kiranchandramohan, the failure may be as a result of compiler-rt not being built. Is this the case with these build bots?

PR #144, opened 2024-07-22
Check the returns of WIFSIGNALED and WIFEXITED

kkwli:
In https://pubs.opengroup.org/onlinepubs/9699919799/functions/wait.html,
- for `WTERMSIG`, 
"If the value of WIFSIGNALED(stat_val) is non-zero, this macro evaluates to the number of the signal that caused the termination of the child process."
- for `WEXITSTATUS`,
"If the value of WIFEXITED(stat_val) is non-zero, this macro evaluates to the low-order 8 bits of the status argument ..."

We need to check `WIFSIGNALED` and `WIFEXTED` before using the values to update `rercode` and `signal`.   On AIX, if `WIFSIGNALED` is zero, `WTERMSIG` has a non-zero value (`0xffffffff`).

### DIFF
--- a/tools/not.cpp
+++ b/tools/not.cpp
@@ -96,7 +96,7 @@ int main(int argc, char* const* argv) {
   // On POSIX systems and Solaris, result is a composite value of the exit code
   // and, potentially, the signal that caused termination of the command.
   retcode = WEXITSTATUS(result);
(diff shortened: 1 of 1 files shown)


tarunprabhu (approved):
LGTM. Thanks for the fix.

PR #145, opened 2024-07-25
[Fortran/gfortran] Remove HLFIR-specific allow and deny lists

tarunprabhu:
The code that supports "feature-specific" testing has been retained, but has been tweaked a little bit. The README has been updated with some examples.


jeanPerier (approved):
Thanks for the clean-up

tblah (approved):
Thanks!

PR #149, opened 2024-08-01
[Fortran/gfortran] Disable test to fix failing buildbots

tarunprabhu:
Several of the AArch64 buildbots are failing due to an undefined reference to __trampoline_setup when compiling internal_dummy_2.f08. This is probably an issue with the buildbot configuration, but in the interest of getting them bots unstuck, the test is being disabled.


antmox (approved):
I agree there is certainly something else that needs to be fixed. Hopefully this will fix some of the current bot issues.

PR #152, opened 2024-08-20
[flang] Disable submodule_3.f08

klausler:
The test Fortran/gfortran/regression/submodule_3.f08 is not expected to compile because it uses basic F'08 features while being compiled by gfortran under -std=f2003.  Flang-new wasn't compiling it either, but that was because it had a bug (also in gfortran) dealing with the inheritance of implicit typing rules in procedure interfaces. That bug is (or shortly will be) fixed, and the test now compiles without error, an unexpected result.  Disable the test.


psteinfeld (approved):
Looks good to me!

tarunprabhu (approved):
Thanks

PR #153, opened 2024-08-20
[Fortran/gfortran] Disable flaky test random_init_2.f90

luporl:
The random_init_2.f90 test is flaky and may fail sometimes, even
when no Fortran/Flang changes are made:
- https://lab.llvm.org/buildbot/#/builders/143/builds/1236
- https://lab.llvm.org/buildbot/#/builders/17/builds/2058

Looking at the test source, it seems there is no guarantee that 2
distinct random numbers will never collide. This is made worse by
multiplying them by 1e6 and converting them to integer, which could
make distinct but close enough single-precision floating-point
numbers give the same result and fail the test.

### DIFF
--- a/Fortran/gfortran/regression/DisabledFiles.cmake
+++ b/Fortran/gfortran/regression/DisabledFiles.cmake
@@ -1747,4 +1747,7 @@ file(GLOB FAILING_FILES CONFIGURE_DEPENDS
   # __trampoline_setup. This is probably an unrelated issue, but as a quick fix
   # for the buildbot, this is disabled.
   internal_dummy_2.f08
+
+  # These are flaky tests, which may fail sometimes.
+  random_init_2.f90
 )


tarunprabhu (approved):
Approving this to get the buildbots going ...

Do you know why this was never triggered until just now? I have never seen this failure in any of my local testing, and none of the buildbots have complained about this in the past either.

luporl:
Thanks for the review!
 
> Do you know why this was never triggered until just now? I have never seen this failure in any of my local testing, and none of the buildbots have complained about this in the past either.

Actually, this test didn't start failing now. It's quite uncommon, but it fails sometimes.
https://lab.llvm.org/buildbot/#/builders/17 failed today, in a single build, and passed in the next one.
The failure in https://lab.llvm.org/buildbot/#/builders/143 is from 17 days ago, also in a single build, with no changes related to Flang.

I suspect it doesn't happen often because most times the generated random numbers, converted to integers, are different. The failure may also go unnoticed, as it usually disappears in the next build.

tarunprabhu:
I see. Thanks for the clarification. There are other `random_*` tests. Do you know if those need to be disabled as well?

luporl:
> I see. Thanks for the clarification. There are other `random_*` tests. Do you know if those need to be disabled as well?

I've checked the `random_init*` ones and they should be fine:
- random_init_1.f90 - expects a compile error
- random_init.f90 - should be fine: generates 5 reals and expects at least 1 to be different than 0.0
- remaining random_init_*.f90 - already disabled: `unimplemented: intrinsic: this_image`

I haven't checked the other `random_*` tests, but I don't remember seeing any failures with those.

DavidSpickett (approved):
Great work investigating Leandro, thanks!

DavidSpickett:
(gonna go ahead and merge this myself)

PR #154, opened 2024-08-22
[Fortran/gfortran][NFC] Recategorize tests

tarunprabhu:
Move some disabled tests to the "unsupported" category since these will never be fixed.

----------------

@lupori, could you check that the inline comment describing the `random_init_2.f90` test is correct.

### DIFF
--- a/Fortran/gfortran/regression/DisabledFiles.cmake
+++ b/Fortran/gfortran/regression/DisabledFiles.cmake
@@ -1824,6 +1824,7 @@ file(GLOB FAILING_FILES CONFIGURE_DEPENDS
   selected_logical_kind_2.f90
   submodule_3.f08
   submodule_33.f08
+  achar_2.f90
   allocate_with_source_30.f90
   allocate_with_source_31.f90
   backslash_1.f90
--- a/Fortran/gfortran/regression/DisabledFiles.cmake
+++ b/Fortran/gfortran/regression/DisabledFiles.cmake
@@ -130,6 +130,11 @@ file(GLOB UNSUPPORTED_FILES CONFIGURE_DEPENDS
   # Test is not conformant as it writes to a constant argument
   # Similar test, that is conformant, added to UnitTests/assign-goto
   assign_5.f90
+
+  # Test is not conformant as it expects different value of cmdstat and cmdmsg
+  # Similar test added: UnitTests/execute_command_line
+  execute_command_line_1.f90
+  execute_command_line_3.f90
 )
 
 # These tests are skipped because they hit a 'not yet implemented' assertion
@@ -301,11 +306,6 @@ file(GLOB UNIMPLEMENTED_FILES CONFIGURE_DEPENDS
 
   # unimplemented: intrinsic: co_broadcast
   coarray_collectives_17.f90
-
-  # Test is not conformant as it expects different value of cmdstat and cmdmsg
-  # Similar test added: UnitTests/execute_command_line
-  execute_command_line_1.f90
-  execute_command_line_3.f90
 
   # unimplemented: intrinsic: failed_images
   coarray_failed_images_1.f08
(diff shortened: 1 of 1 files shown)


Fortran/gfortran/regression/DisabledFiles.cmake
```diff
@@ -119,6 +119,22 @@ file(GLOB UNSUPPORTED_FILES CONFIGURE_DEPENDS
   unlimited_polymorphic_14.f90
   # Unsupported predefined macro: __TIMESTAMP__
   wdate-time.F90
+
+  # This test occasionally fails. It checks that two arrays initialized with
+  # random numbers are not identical. This might be the case in gfortran, but in
+  # flang, it is possible, though unlikely, for two "randomly initialized"
+  # arrays to be identical. Unless something in flang's implementation of the
+  # RANDOM_* intrinsics changes, this test will be unsupported.
```
luporl, line 127: > It checks that two arrays initialized with random numbers are not identical.

While this description isn't wrong, it's slightly incomplete/inaccurate.

First, the test initializes two arrays from random real numbers (in the range 0-1), multiplied by 1e6, converted to integer and back to real. Some precision may be lost in this process, as only 6 digits are preserved.
Then the test checks if any element in the first array matches the corresponding element in the second array, aborting if true.
Because of the precision loss and random nature of the test, 2 elements may end up matching.

This is the first part of the test. The second part initializes the arrays again, but this time using a repeatable pseudo random number generator, and then checks if the arrays are equal, aborting if not. This part looks correct to me and can be omitted in the comment.

Maybe something like:
```
  # This test occasionally fails. It checks that two arrays initialized with
  # random real numbers, converted to integers, are not identical. This
  # might be the case in gfortran, but in flang, because of the precision
  # loss and random nature of the test, it is possible, though unlikely, for
  # two "randomly initialized" array elements to be identical. Unless
  # something in flang's implementation of the RANDOM_* intrinsics
  # changes, this test will be unsupported.
```

Although I guess this test is probably flaky with gfortran too.

tblah: I wonder if flang's random number generation has some bias that is leading to producing the same numbers more often than gfortran. Even 6 digits should make it rare for these to fail.

But +1 for disabling it. This isn't a good test design because it will fail spuriously with small likelihood.

jeanPerier: I think flang may lack some randomness here, I opened https://github.com/llvm/llvm-project/issues/106221 with some more statistically relevant facts (the gfortran test should indeed do a bit of sampling, even gfortran will fail it once in 100000).

tarunprabhu: Yes, I am not too keen on enabling this myself. We could consider adding a similar test of our own, but even then, I am not sure how to (or if we even should) test for "randomness".

jeanPerier (approved):
Thanks, LGTM

Fortran/gfortran/regression/DisabledFiles.cmake
```diff
@@ -119,6 +119,23 @@ file(GLOB UNSUPPORTED_FILES CONFIGURE_DEPENDS
   unlimited_polymorphic_14.f90
   # Unsupported predefined macro: __TIMESTAMP__
   wdate-time.F90
+
+  # This test occasionally fails. It checks that two arrays initialized with
+  # random real numbers, converted to integers, are not identical. This might be
+  # the case in gfortran, but in flang, because of the precision loss and random
+  # nature of the test, it is possible, for two "randomly initialized" arrays to
+  # be identical. Unless something in flang's implementation of the RANDOM_*
+  # intrinsics changes, this test will be unsupported.
```
tblah, line 128: nit: this should be fixed soon in flang and so the wording should change here to make it clear there isn't a bug in flang compared to gfortran, the test is just inherently unreliable

See https://github.com/llvm/llvm-project/pull/106250

tblah (approved):

Fortran/gfortran/regression/DisabledFiles.cmake
tblah: LGTM, thanks

luporl: LGTM.

PR #156, opened 2024-09-05
[Fortran] Disable pr32601.f03

clementval:
https://github.com/llvm/llvm-project/pull/107353 will relax the check when -pedantic is not enabled. `-std=f2003` is not supported in flang. Disable the test until we have a way to add flags.

### DIFF
--- a/Fortran/gfortran/regression/DisabledFiles.cmake
+++ b/Fortran/gfortran/regression/DisabledFiles.cmake
@@ -1840,4 +1840,7 @@ file(GLOB FAILING_FILES CONFIGURE_DEPENDS
   pr111022.f90
   pr114304.f90
   zero_sized_15.f90
+
+  # Test needs to add -pedantic to show the error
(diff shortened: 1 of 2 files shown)


tarunprabhu:
We don't allow changes to the actual tests in this suite. Trying to find a way to add/remove flags so such tests can be run is on my list of things to do. 

For now, there are two options:

1) Disable this test. You can add it to the FAILING section of `Fortran/gfortran/regression/DisabledFiles.cmake` with a comment indicating why it has been disabled. 

2) Disable this test and add a similar one to `Fortran/UnitTests`. There are a couple of examples of this such as `Fortran/UnitTests/assign-goto`.

tarunprabhu (changes requested):
See my previous comment on what needs to be changed. Thanks.

tarunprabhu (approved):
Thanks for making the change, Valentin. :-)

PR #157, opened 2024-09-06
[SPEC] Workaround error with latest clang in 510.parest_r

lukel97:
Similarly to #118, 510.parest_r contains another latent error related to template deduction that after https://github.com/llvm/llvm-project/pull/100692 fails to build. We can workaround this by passing -fno-relaxed-template-template-args


mizvekov:
We have just merged into clang the core defect fix for this issue, this patch should not be necessary anymore.

lukel97:
> We have just merged into clang the core defect fix for this issue, this patch should not be necessary anymore.

That's great to hear, thanks.

PR #159, opened 2024-09-09
Fix spurious errors in halide tests

nikic:
We occasionally see spurious llvm-test-suite failures with the following error message:

    CMake Error: failed to create symbolic link '/tmp/tmp.x1PxF3OhCA/Bitcode/Benchmarks/Halide/local_laplacian/../images/rgb.bytes': File exists

The reason for this is that two halide tests try to copy a test input to the same location, and create_symlink can fail if these calls race.

Fix this by putting the inputs for the tests into different directories, by not making ".." part of the result directory.


fhahn (approved):
LGTM, thanks!

PR #160, opened 2024-09-09
fpcmp: Dump inputs on comparison failure

nikic:
There are occasionally spurious failures in llvm-test-suite, and you end up with a phenomenally unhelpful message like:

    tools/fpcmp-target: Comparison failed, textual difference between '0' and '7'

This is good enough if you can inspect the inputs to fpcmp locally, but unhelpful for CI failures.

Improve things by dumping the inputs to fpcmp on a textual comparison failure.

### DIFF
--- a/tools/fpcmp.c
+++ b/tools/fpcmp.c
@@ -247,8 +247,25 @@ char *load_file(const char *path, long *size_out) {
   return data;
 }
 
-void dump_inputs(const char *data_a, const char *data_b) {
-    fprintf(stderr, "\nInput 1:\n%s\nInput 2:\n%s", data_a, data_b);
+static bool contains_non_printable_characters(const char *data) {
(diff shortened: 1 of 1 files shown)


MatzeB:
I think some benchmarks may spit out binary data where printing it all out on the terminsl may not be nice… Maybe there is some easy heuristic like not printing when there are non-printable characters in the output…

nikic:
> I think some benchmarks may spit out binary data where printing it all out on the terminsl may not be nice… Maybe there is some easy heuristic like not printing when there are non-printable characters in the output…

That's a good point! I've adjusted the implementation to not dump the contents for non-printable chars. I guess an alternative would be to use a hex dump in that case.

MatzeB (approved):
LGTM

PR #161, opened 2024-09-10
[flang] Disable two tests as error is being downgraded

klausler:
Fortran/gfortran/regression/restricted_expression_2.f90 and Fortran/gfortran/regression/restricted_expression_3.f90 are disabled with this patch, as the condition that they test will soon no longer be an error with flang, just a warning. (This is the use of a local variable with proper initialization and the SAVE attribute being used as part of a specification expression.)


tarunprabhu (approved):
Thanks.

PR #163, opened 2024-09-11
[flang] Disable test that expects error from new extension

klausler:
I'm shortly adding the ability to allow real-looking input for NAMELIST integer values.  GNU Fortran doesn't allow that, and has a test to check its error detection, so disable that test.


tarunprabhu (approved):
Thanks.

PR #165, opened 2024-09-24
[MicroBenchmarks/LoopVectorization] Add Microbenchmark for Epilogue Vectorization

juliannagele:
This microbenchmark attempts to test the impact of epilogue vectorization on loops with varying epilogue lengths and vector widths for loops with and without reductions.

### DIFF
--- a/MicroBenchmarks/LoopVectorization/EpilogueVectorization.cpp
+++ b/MicroBenchmarks/LoopVectorization/EpilogueVectorization.cpp
@@ -65,9 +65,18 @@ template <typename Ty> void benchReductionAutoVec(benchmark::State &state) {
   runBenchForEpilogueVectorization<Ty>(state, &loopWithReductionAutoVec<Ty>);
 }
 
+#ifdef ALL_LOOP_EPILOGUE_TESTS
 BENCHMARK_TEMPLATE(benchAutoVec, uint8_t)->DenseRange(65, 127, 1);
 BENCHMARK_TEMPLATE(benchReductionAutoVec, uint8_t)->DenseRange(65, 127, 1);
 BENCHMARK_TEMPLATE(benchAutoVec, uint16_t)->DenseRange(65, 127, 1);
 BENCHMARK_TEMPLATE(benchReductionAutoVec, uint16_t)->DenseRange(65, 127, 1);
 BENCHMARK_TEMPLATE(benchAutoVec, uint32_t)->DenseRange(65, 127, 1);
 BENCHMARK_TEMPLATE(benchReductionAutoVec, uint32_t)->DenseRange(65, 127, 1);
+#else
+BENCHMARK_TEMPLATE(benchAutoVec, uint8_t)->Range(65, 127);
+BENCHMARK_TEMPLATE(benchReductionAutoVec, uint8_t)->Range(65, 127);
+BENCHMARK_TEMPLATE(benchAutoVec, uint16_t)->Range(65, 127);
+BENCHMARK_TEMPLATE(benchReductionAutoVec, uint16_t)->Range(65, 127);
+BENCHMARK_TEMPLATE(benchAutoVec, uint32_t)->Range(65, 127);
+BENCHMARK_TEMPLATE(benchReductionAutoVec, uint32_t)->Range(65, 127);
+#endif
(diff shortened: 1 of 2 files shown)


MicroBenchmarks/LoopVectorization/EpilogueVectorization.cpp
```diff
@@ -0,0 +1,89 @@
...
+    runBenchForEpilogueVectorization<Ty>(state, &loopAutoVec<Ty>, Itr);        \
+  }                                                                            \
+  BENCHMARK(benchAutoVecFor##Ty##ForLoopTC##Itr);                              \
+  void benchReductionAutoVecFor##Ty##ForLoopTC##Itr(benchmark::State &state) { \
+    runBenchForEpilogueVectorization<Ty>(state, &loopWithReductionAutoVec<Ty>, \
+                                         Itr);                                 \
+  }                                                                            \
+  BENCHMARK(benchReductionAutoVecFor##Ty##ForLoopTC##Itr);
+
+ADD_BENCHMARK(uint8_t, 65)
```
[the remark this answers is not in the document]
fhahn: It might be worth covering all possible trip counts for the epilogue, using something like

```
BENCHMARK_TEMPLATE(uint8_t)->DenseRange(65, 83, 1);
```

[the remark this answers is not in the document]
juliannagele: Ah, neat, thanks, I was wondering if there's a better way; now using `BENCHMARK_TEMPLATE(...)->DenseRange(...)` to get all trip counts.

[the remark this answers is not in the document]
juliannagele: Now also running only a small set of tests by default to avoid potentially overloading build bots that run the test suite.

MicroBenchmarks/LoopVectorization/EpilogueVectorization.cpp
```diff
@@ -0,0 +1,82 @@
...
+  }
+  return sum;
+}
+
+template <typename Ty> void benchAutoVec(benchmark::State &state) {
+  runBenchForEpilogueVectorization<Ty>(state, &loopAutoVec<Ty>);
+}
+
+template <typename Ty> void benchReductionAutoVec(benchmark::State &state) {
+  runBenchForEpilogueVectorization<Ty>(state, &loopWithReductionAutoVec<Ty>);
```
fhahn, line 65: To address @sjoerdmeijer, you could use `callThroughOptnone` like in `MicroBenchmarks/LoopVectorization/RuntimeChecks.cpp`, if it doesn't impact the benchmarks negatively?

MicroBenchmarks/LoopVectorization/EpilogueVectorization.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+#ifdef ALL_LOOP_EPILOGUE_TESTS
+BENCHMARK_TEMPLATE(benchAutoVec, uint8_t)->DenseRange(65, 127, 1);
+BENCHMARK_TEMPLATE(benchReductionAutoVec, uint8_t)->DenseRange(65, 127, 1);
+BENCHMARK_TEMPLATE(benchAutoVec, uint16_t)->DenseRange(65, 127, 1);
+BENCHMARK_TEMPLATE(benchReductionAutoVec, uint16_t)->DenseRange(65, 127, 1);
+BENCHMARK_TEMPLATE(benchAutoVec, uint32_t)->DenseRange(65, 127, 1);
+BENCHMARK_TEMPLATE(benchReductionAutoVec, uint32_t)->DenseRange(65, 127, 1);
+#else
+BENCHMARK_TEMPLATE(benchAutoVec, uint8_t)->Range(65, 127);
```
fhahn, line 76: What's the default step? Might be good to document

MicroBenchmarks/LoopVectorization/EpilogueVectorization.cpp
juliannagele: Thanks! Using `callThroughOptnone` all benchmarks are ever so slightly slower but behave the same otherwise -- adopted.

MicroBenchmarks/LoopVectorization/EpilogueVectorization.cpp
juliannagele: Ah, yes looks like it's actually a *8 multiplier by default, which is a bit silly here, changed it to using `->Arg(...)` to be explicit on what is run.

fhahn (approved):
LGTM, thanks!

PR #169, opened 2024-10-04
[Fortran] disbale failing tests after assumed-rank were enabled

jeanPerier:
Fix https://lab.llvm.org/buildbot/#/builders/143/builds/2600

Some tests expecting a semantic error wrongly passed because of the lowering TODO for assumed rank that was enabled by default in https://github.com/llvm/llvm-project/pull/110893.


jeanPerier:
Opened https://github.com/llvm/llvm-project/issues/111120 to triage the failing tests.

PR #171, opened 2024-10-11
[flang] Disable two xfail tests that will soon not fail

klausler:
An upcoming PR to flang (https://github.com/llvm/llvm-project/pull/112024) will soon allow two gfortran tests to compile and run successfully, which will come as a fatal surprise because they're marked "xfail".

### DIFF
--- a/Fortran/gfortran/regression/DisabledFiles.cmake
+++ b/Fortran/gfortran/regression/DisabledFiles.cmake
@@ -1850,4 +1850,8 @@ file(GLOB FAILING_FILES CONFIGURE_DEPENDS
 
   # Tests expect semantic errors that are not raised.
   c_sizeof_7.f90
+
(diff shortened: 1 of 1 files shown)


tarunprabhu (changes requested):
The `tests.cmake` file is auto-generated and must not be edited by hand because it will be overwritten the next time the test suite is sync'ed with the upstream tests. 

I am working on a way to override the DejaGNU annotations so that we can use such tests where the behavior of gfortran and flang are different. But for now, they would have to be disabled. If you could disable them and leave a comment indicating that they should not be xfail, I can re-enable them when the overriding mechanism is in place.

Thanks.

klausler:
Updated.

tarunprabhu (approved):
Thanks for making the change.

PR #174, opened 2024-10-22
[LoongArch] Add vector floating-point check function

wangleiat:
Because vector support for constant folding was added in visitFMA, the [x]vfmadd.d instructions was optimized into a constant load, causing a bitwise mismatch for NaN values between the expected and actual results. (The NaN from constant folding differs from the NaN generated by the
 instruction.)

https://lab.llvm.org/staging/#/builders/20/builds/4041


wangleiat:
@leecheechen

SixWeining (approved):

PR #180, opened 2024-11-20
[cmake] Support more Fortran filename suffixes

kawashima-fj:
These Fortran suffixes are aligned with [the `lookupTypeForExtension` function in `clang/lib/Driver/Types.cpp`](https://github.com/llvm/llvm-project/blob/4d6a5fc/clang/lib/Driver/Types.cpp#L299).

We Fujitsu want to contribute the [Fujitsu Compiler Test Suite](https://github.com/fujitsu/compiler-test-suite/) to the LLVM test-suite after it becomes ready and we obtain consensus in the community. It includes `*.f95` and `*.for` tests.

This commit affects tests which use `llvm_singlesource` or `llvm_multisource` in `CMakeLists.txt`. I confirmed this commit does not affect any existing tests in the LLVM test-suite.

### DIFF
--- a/cmake/modules/SingleMultiSource.cmake
+++ b/cmake/modules/SingleMultiSource.cmake
@@ -33,7 +33,9 @@ function(llvm_singlesource)
   if(DEFINED Source)
     set(sources ${Source})
   else()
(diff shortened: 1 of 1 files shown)


tarunprabhu (approved):
Thanks for working on this. LGTM.

PR #181, opened 2024-11-27
[Fortran/gfortran] Enable gfortran build in LoongArch.

ylzsx:


tarunprabhu (approved):
Thanks for this. 

Right now, there are tests that are disabled on all platforms because they fail on either X86-64 and AArch64 (the only two that we regularly test). We are [in the process](https://github.com/llvm/llvm-test-suite/pull/176) of setting up a framework which will allow such tests to be enabled conditionally on certain platforms. If this results in a failure on LoongArch, let me know and I can disable the test on just LoongArch.

ylzsx:
> Right now, there are tests that are disabled on all platforms because they fail on either X86-64 and AArch64 (the only two that we regularly test). We are [in the process](https://github.com/llvm/llvm-test-suite/pull/176) of setting up a framework which will allow such tests to be enabled conditionally on certain platforms. If this results in a failure on LoongArch, let me know and I can disable the test on just LoongArch.

Thanks. After applying the patch you mentioned, I tested it on a LoongArch64 machine, and it works well.

PR #183, opened 2024-12-02
[fortran] disable non conformant gfortran aliasing_array_result_1 test

jeanPerier:
Needed after https://github.com/llvm/llvm-project/pull/118070

Both ifx and nvfortran/classic flang also "fail" this non conformant test.


tarunprabhu (approved):
Thanks for the description of why it is being "permanently" disabled. 

LGTM

klausler (approved):

PR #184, opened 2024-12-04
Fix tools tests on Windows

DavidTruby:
The commit 89fcc5e introduced some tests for the tools in the tools/ subdirectory that are always run, but these tests were not working on Windows due to the executables ending in .exe and one of the tests being written in bash.

This commit fixes those issues by directly using the $<TARGET_FILE> generator expressions to get the executable file names, and rewriting the bash test in python.

### DIFF
--- a/tools/test/CMakeLists.txt
+++ b/tools/test/CMakeLists.txt
@@ -1,22 +1,27 @@
 # Copy these files to the build directory so that the tests can be run even
 # without the source directory.
-configure_file(test_not.sh test_not.sh
-  @ONLY)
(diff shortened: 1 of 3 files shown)


tarunprabhu (approved):
Thanks for this, David. I'll keep some of these in mind for the future. I am afraid Windows compatibility completely slipped my mind when I added these.

DavidTruby:
No worries, thanks for the quick review! I'm more alarmed that nobody noticed this was broken on Windows until I ran it manually recently than I am that it broke in the first place; are there no bots running these tests on Windows regularly? 😓

PR #185, opened 2024-12-04
Always pass NO_STOP_MESSAGE to Fortran tests

DavidTruby:
Currently out of the box the Fortran tests fail with flang because the NO_STOP_MESSAGE environment variable needs to be set for them to pass. This environment variable should therefore always be set for the tests so that they can pass out of the box.

### DIFF
--- a/Fortran/CMakeLists.txt
+++ b/Fortran/CMakeLists.txt
@@ -17,3 +17,5 @@ if (NOT WIN32 AND NOT APPLE)
     add_subdirectory(gfortran)
   endif()
 endif()
+
+file(COPY lit.local.cfg DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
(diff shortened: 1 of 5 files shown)


tarunprabhu (commented):
Thanks. This should remove a common stumbling block for the tests. 

Could we document that this environment variable is set automatically when running the test suite? If the default behavior of flang is to print a message, we may want to state explicitly that the tests here will squelch that. It looks like there isn't a top-level readme in `Fortran/`, but putting it there should be sufficient.

Do you think there is a use case for _not_ setting `NO_STOP_MESSAGE`?

tarunprabhu:
> > Do you think there is a use case for not setting NO_STOP_MESSAGE?
> 
> The only case I see is for tests that would be checking the STOP message, it is a limited use case and these tests would not be portable given stop message texts are compiler dependent. If ever added, such tests could be in their own directory where the lit cfg could remove the "NO_STOP_MESSAGE".

This sounds reasonable. With some documentation about the environment variable, this should be good to go.

DavidSpickett:
If you have the time, removing NO_STOP_MESSAGE from the bot configurations in llvm-zorg would be a nice cleanup after this.

tarunprabhu (approved):
Thanks David.

PR #186, opened 2024-12-07
Add a placeholder for the test-suite-externals directory

mstorsjo:
When setting up external sources within llvm-test-suite, that can either be done by checking out those sources in the test-suite-externals directory, or by setting a CMake variable to point at the source directory elsewhere.

As the git repo itself doesn't contain any files in the test-suite-externals directory, no such directory exists in users checkouts, unless they manually create it (which feels brittle and can be prone to mistypings etc).

Therefore, add a placeholder dummy file to force git to create the directory, making it more obvious to users where they can place external sources. The directory is ignored via .gitignore, so files or directories in this tree don't get added accidentally.


sjoerdmeijer (approved):
This makes a lot of sense, I think this is an improvement, so LGTM.

fhahn:
Won't this break checkouts for users already having a test-suite-externals directory in their clone?

mstorsjo:
> Won't this break checkouts for users already having a test-suite-externals directory in their clone?

No - this doesn't come with an explicit intent of "create a new directory", it'll just add the `.gitkeep` file in the existing directory, or create the directory if needed.

fhahn (approved):
LGTM, thanks for clarifying

PR #190, opened 2024-12-12
Avoid UB in huffbench

nikic:
The `--heap` results in an out of bounds pointer index. Fix this by subtracting 1 during indexing instead.


serge-sans-paille:
Looks good, funny to see a benchmark having some UB ^^!

serge-sans-paille:
I have access to a machine where I was able to reproduce the issue, let me confirm your patch works.

serge-sans-paille:
works like a charm!

serge-sans-paille (approved):

PR #187, opened 2024-12-09
[Fortran/gfortran] Disable tests only on AArch64, but enable them els…

tarunprabhu:
…ewhere

Some tests fail only on AArch64, but pass on other platforms. Previously, these were disabled everywhere, but now they can be disabled only on AArch64 and enabled on other platforms. In addition, some tests that were failing only on AArch64 now pass, so they are enabled everywhere.

The static test configuration has also been updated to reflect these changes. Some bugs were exposed in the update script during this process. These have also been fixed.

----------------------------------------------------------------------------------------
@ylzsx, 

Since these tests were preciously disabled on LoongArch, could you check that they pass there. If they do not, I will update the PR to disable them on LoongArch as well.

PR #191, opened 2024-12-12
[Fortran/gfortran] Use -O0 for cshift_bounds_[34].

vzakhari:
There is no support for `-fbounds-check` in Flang yet.
There is a related feature request at https://github.com/orgs/llvm/projects/12?pane=issue&itemId=29048733
Until then, the optimizations may replace the Fortran runtime calls
(which usually provide all the error checking) with inline sequences
that do not check for errors to provide the fast code
(e.g. https://github.com/llvm/llvm-project/pull/119480).
For the time being I would like to disable these tests.


tarunprabhu:
The `tests.cmake` file is auto-generated and not meant to be edited by hand. Support to override some gfortran attributes was [recently added](https://github.com/llvm/llvm-test-suite/pull/176). The first PR to make use of this is [under review](https://github.com/llvm/llvm-test-suite/pull/187). This could be used to do what you want. Right now, there is no support to override command-line options as you have requested, but it is next on my list of things to do. 

For now, if these are causing failures, you could disable the tests by adding them to `Fortran/gfortran/regression/DisabledFiles.cmake`. If you could leave a comment indicating that they can be enabled by passing `-O0`, I will know to enable them when I add support for overriding command-line options.

vzakhari:
Thank you, Tarun! I will disable the tests with a comment.

tarunprabhu (approved):
Thanks for the change, Slava.

PR #192, opened 2024-12-16
[Fortran/gfortran] Enable tests that now pass

tarunprabhu:
Enable a number of tests that were disabled, but now pass.

----------------------------------------
Notes for reviewers

In the past, some tests pass for me, but fail on different configurations. I have included a larger number of reviewers this time to cover several targets and build/test configurations. If any tests fail on your setup, let me know and I will disable them again. 

As a result of enabling these tests, some of the `DisabledFiles.cmake` files are empty. I have chosen to retain the files for now because there have been instances of regressions requiring tests to be temporarily disabled. Once `flang` becomes more stable, we can consider removing them.

### DIFF
--- a/Fortran/gfortran/regression/DisabledFiles.cmake
+++ b/Fortran/gfortran/regression/DisabledFiles.cmake
@@ -919,6 +919,7 @@ file(GLOB FAILING_FILES CONFIGURE_DEPENDS
   unformatted_subrecord_1.f90
   widechar_IO_4.f90
   zero_sized_1.f90
+  do_check_1.f90
   random_3.f90
 
   # ---------------------------------------------------------------------------
--- a/Fortran/gfortran/regression/DisabledFiles.cmake
+++ b/Fortran/gfortran/regression/DisabledFiles.cmake
@@ -148,52 +148,14 @@ file(GLOB UNSUPPORTED_FILES CONFIGURE_DEPENDS
 file(GLOB UNIMPLEMENTED_FILES CONFIGURE_DEPENDS
   # unimplemented: assumed rank in procedure interface.
   ISO_Fortran_binding_1.f90
-  ISO_Fortran_binding_13.f90
   ISO_Fortran_binding_3.f90
-  PR100029.f90
-  PR100097.f90
-  PR100098.f90
   PR100911.f90
   PR100915.f90
-  PR93963.f90
-  PR94022.f90
-  PR95196.f90
-  PR95352.f90
-  PR96726.f90
-  PR96727.f90
-  PR96728.f90
-  PR97046.f90
-  assumed_rank_12.f90
-  assumed_rank_18.f90
-  assumed_rank_19.f90
-  assumed_rank_20.f90
-  assumed_rank_21.f90
-  assumed_type_9.f90
-  assumed_type_10.f90
-  assumed_type_11.f90
-  bind-c-contiguous-2.f90
   interface_49.f90
-  is_contiguous_2.f90
-  pr103366.f90
-  pr84088.f90
-  pr88932.f90
-  pr92277.f90
-  pr95828.f90
-  select_rank_5.f90
-  sizeof_4.f90
   sizeof_6.f90
   unlimited_polymorphic_1.f03
-  unlimited_polymorphic_32.f90
-
-  # unimplemented: assumed-rank variable in procedure implemented in Fortran
-  associate_66.f90
-  bind_c_optional-2.f90
-  intent_out_19.f90
-  intent_out_20.f90
-  shape_12.f90
 
   # unimplemented: ASYNCHRONOUS in procedure interface
-  assumed_rank_13.f90
   asynchronous_3.f03
 
   # unimplemented: assumed type in actual argument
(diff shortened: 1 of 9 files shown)


tblah (approved):
Passes for me on aarch64 with optimization enabled. Thanks for keeping on top of this!

Fortran/gfortran/regression/DisabledFiles.cmake
```diff
@@ -1010,16 +919,8 @@ file(GLOB FAILING_FILES CONFIGURE_DEPENDS
   unformatted_subrecord_1.f90
   widechar_IO_4.f90
   zero_sized_1.f90
-  elemental_function_2.f90
-  do_check_1.f90
```
kiranchandramohan, line 1014: This failed for me.

Fortran/gfortran/regression/gomp/DisabledFiles.cmake
```diff
@@ -28,7 +28,6 @@ file(GLOB UNIMPLEMENTED_FILES CONFIGURE_DEPENDS
 
   # unimplemented: support for polymorphic types
   pr102621.f90
-  pr86470.f90
```
tblah, line 31: On closer inspection, this shouldn't pass. The privatizer is incorrect:
```
  omp.private {type = private} @_QFEval_private_ref_class_heap_none : !fir.ref<!fir.class<!fir.heap<none>>> alloc {
  ^bb0(%arg0: !fir.ref<!fir.class<!fir.heap<none>>>):
    %0 = fir.alloca !fir.class<!fir.heap<none>> {bindc_name = "val", pinned, uniq_name = "_QFEval"}
    %1 = fir.load %arg0 : !fir.ref<!fir.class<!fir.heap<none>>>
    %2 = fir.box_addr %1 : (!fir.class<!fir.heap<none>>) -> !fir.heap<none>
    %3 = fir.convert %2 : (!fir.heap<none>) -> i64
    %c0_i64 = arith.constant 0 : i64
    %4 = arith.cmpi ne, %3, %c0_i64 : i64
    fir.if %4 {
      %6 = fir.load %arg0 : !fir.ref<!fir.class<!fir.heap<none>>>
      %7 = fir.allocmem none {fir.must_be_heap = true, uniq_name = "_QFEval.alloc"}
      %8 = fir.embox %7 : (!fir.heap<none>) -> !fir.class<!fir.heap<none>>
      fir.store %8 to %0 : !fir.ref<!fir.class<!fir.heap<none>>>
    } else {
      %6 = fir.zero_bits !fir.heap<none>
      %7 = fir.embox %6 : (!fir.heap<none>) -> !fir.class<!fir.heap<none>>
      fir.store %7 to %0 : !fir.ref<!fir.class<!fir.heap<none>>>
    }
    %5:2 = hlfir.declare %0 {fortran_attrs = #fir.var_attrs<allocatable>, uniq_name = "_QFEval"} : (!fir.ref<!fir.class<!fir.heap<none>>>) -> (!fir.ref<!fir.class<!fir.heap<none>>>, !fir.ref<!fir.class<!fir.heap<none>>>)
    omp.yield(%5#0 : !fir.ref<!fir.class<!fir.heap<none>>>)
  } dealloc {
  ^bb0(%arg0: !fir.ref<!fir.class<!fir.heap<none>>>):
    %0 = fir.load %arg0 : !fir.ref<!fir.class<!fir.heap<none>>>
    %1 = fir.box_addr %0 : (!fir.class<!fir.heap<none>>) -> !fir.heap<none>
    %2 = fir.convert %1 : (!fir.heap<none>) -> i64
    %c0_i64 = arith.constant 0 : i64
    %3 = arith.cmpi ne, %2, %c0_i64 : i64
    fir.if %3 {
      %false = arith.constant false
      %4 = fir.absent !fir.box<none>
      %5 = fir.address_of(@_QQclX7c9724ff976eac349f8980ba67cae772) : !fir.ref<!fir.char<1,75>>
      %c75 = arith.constant 75 : index
      %c6_i32 = arith.constant 6 : i32
      %6 = fir.zero_bits !fir.ref<none>
      %7 = fir.convert %arg0 : (!fir.ref<!fir.class<!fir.heap<none>>>) -> !fir.ref<!fir.box<none>>
      %8 = fir.convert %5 : (!fir.ref<!fir.char<1,75>>) -> !fir.ref<i8>
      %9 = fir.call @_FortranAAllocatableDeallocatePolymorphic(%7, %6, %false, %4, %8, %c6_i32) fastmath<contract> : (!fir.ref<!fir.box<none>>, !fir.ref<none>, i1, !fir.box<none>, !fir.ref<i8>, i32) -> i32
    }
    omp.yield
  }
```

The polymorphic type should be allocated using the runtime not with a malloc of `none` (which I think is `malloc(0)`).

This test happens to pass because the mold variable is not allocated when entering the parallel region so the branch that tries to do this allocation is never reached.

I have some downstream changes to privatization which will correctly hit a TODO on this test instead of generating incorrect code. Please don't enable this so that it doesn't have to be disabled again when I finish my changes.

tblah (changes requested):

tarunprabhu:
Thanks @tblah and @kiranchandramohan for taking a look. The tests that were incorrectly passing/failing have been disabled.

jeanPerier (approved):
Thank you Tarun, this passes with my build and, for the few tests I checked, is in line with my expectations.

kiranchandramohan (approved):
LG. The latest set passes for me.

ylzsx (approved):
Thank you. Passes for me on LoongArch64 with optimization `-O2` enabled.

DavidTruby (approved):
LGTM. 

All the gfortran tests are disabled on Windows so I haven't checked any of this there; some day we should go through and enable any of the ones that should work on Windows but that's not really relevant to this PR.

tarunprabhu:
> All the gfortran tests are disabled on Windows so I haven't checked any of this there; some day we should go through and enable any of the ones that should work on Windows but that's not really relevant to this PR.

Ah, right. I think the main reason I had them disabled was because some of the tests use the `diff` utility on *nix. I think I added some code to use `cmp` on Windows, but since I had no way to test it, I just disabled the entire thing on Windows. 

Let me know when you have some time to look into this. It would be good to have this running on Windows, especially since it looks like these will be enabled on more architectures as well.

AlexisPerry (approved):
All passes for my configuration (x86, amd-rome node).

PR #193, opened 2024-12-20
[SingleSource/Vectorizer] Add unit tests for FindLastIV pattern.

Mel-Chen:
This patch adds runtime test case for vectorization of FindLastIV reduction idiom:
```
int32_t Rdx = -1;
for (int32_t I = 0; I < TC; I++) {
  Rdx = A[I] > B[I] ? I : Rdx;
}
return Rdx;
```
Improving test coverage for https://github.com/llvm/llvm-project/pull/67812 and https://github.com/llvm/llvm-project/pull/120395.


alexey-bataev (approved):
Looks good to me, but wait for couple days for others to review

Mel-Chen:
@fhahn @Meinersbur Do you have any suggestions and comments?

fhahn (commented):
Same as for, do we need smaller types to trigger epilogue vectorization on more platforms?

IIUC https://github.com/llvm/llvm-project/pull/120395 will need to be fixed first, otherwise the new tests will fail?

Mel-Chen:
> Same as for, do we need smaller types to trigger epilogue vectorization on more platforms?

Sure, could you take a look at AnyOf https://github.com/llvm/llvm-test-suite/pull/195 to see if the approach is appropriate? I will modify this patch in the same way if you feel good.
> 
> IIUC [llvm/llvm-project#120395](https://github.com/llvm/llvm-project/pull/120395) will need to be fixed first, otherwise the new tests will fail?

Yes, the current implementation still produces incorrect output in the following datasets after enabling epilogue vectorization. 
```
    // Check with only Src1[N - 1] > Src2[N - 1].
    for (unsigned I = 0; I != N; ++I)
      Src1[I] = Src2[I] = std::numeric_limits<Ty>::min();
    Src1[N - 1] = std::numeric_limits<Ty>::max();
```
[llvm/llvm-project#120395](https://github.com/llvm/llvm-project/pull/120395) can fix this error.

SingleSource/UnitTests/Vectorizer/find-last.reference_output
```diff
@@ -0,0 +1,9 @@
+Checking findlast_true_update
+Checking findlast_false_update
+Checking findlast_start_TC
+Checking findlast_inc_2
+Checking findlast_start_decreasing_induction
+Checking findlast_iv_start_3
+Checking findlast_start_3_iv_start_3
+Checking findlast_start_2_iv_start_3
+Checking findlast_start_4_iv_start_3
```
fhahn, line 9: I think this is missing an `exit 0` at the end for the test to pass?

Mel-Chen: 803fa53317abf7c7fcc99c923cf15835efb820b2
Fixed, thanks.

Mel-Chen:
Add test for FFindLastIV, and s16 type for triggering epilogue vectorization easier.
Please take a look again, thanks.

fhahn (approved):
LGTM, thanks!

AaronBallman:
FYI, but this test fails with GCC 15.x because they enabled `-D_GLIBCXX_ASSERTIONS=1` libstdc++ by default. Here's a reproducer on godbolt: https://godbolt.org/z/qh4EGo3s7

We ran into this with our downstream at Intel because we also enable `_GLIBCXX_ASSERTIONS`, but upstream Clang does not currently do that nor does libc++ currently have the same assertions, so that's likely why we've not started hitting this failure in upstream's tree yet.

PR #194, opened 2025-01-02
[Fortran/gfortran] Add minimum python version

kkwli:


tarunprabhu (approved):
Thanks for checking and the fix. LGTM.

PR #195, opened 2025-01-03
[SingleSource/Vectorizer] Add unit tests for AnyOf pattern.

Mel-Chen:
This patch adds runtime test case for vectorization of AnyOf reduction idiom:
```
int32_t Rdx = -1;
for (unsigned I = 0; I < TC; I++) {
  Rdx = A[I] > B[I] ? 3 : Rdx;
}
return Rdx;
```
Improving test coverage, and please refer the issue https://github.com/llvm/llvm-project/issues/120405.


SingleSource/UnitTests/Vectorizer/any-of.cpp
```diff
@@ -0,0 +1,181 @@
...
+  {
+    // Update Rdx to 3 once any A[i] > B[i] is true.
+    DEFINE_SCALAR_AND_VECTOR_FN2(
+        uint32_t Rdx = -1;,
+        for (unsigned I = 0; I < TC; I++) {
+          Rdx = A[I] > B[I] ? 3 : Rdx;
+        }
+        return Rdx;
+        );
+    checkVectorFunction<uint32_t, int32_t>(ScalarFn, VectorFn,
```
fhahn, line 127: It would probably be good to also check with a smaller type (int8_t or int16_t ?) to trigger epilogue vectorization on more platforms?

Mel-Chen: 37b1efc0c9e80356effe9192ef9cf3d960d6f9d0
Yes, I patched it, but only for IAnyOf.
Checking FAnyOf will require an fp type with a smaller size than float to be possible.

Mel-Chen:
Ping

SingleSource/UnitTests/Vectorizer/any-of.reference_output
```diff
@@ -0,0 +1,12 @@
...
+Checking anyof_icmp_s16_true_update
+Checking anyof_icmp_s32_false_update
+Checking anyof_fcmp_s32_false_update
+Checking anyof_icmp_s16_false_update
+Checking anyof_icmp_u32_start_TC
+Checking anyof_fcmp_u32_start_TC
+Checking anyof_icmp_u16_start_TC
+Checking anyof_icmp_u32_update_by_TC
+Checking anyof_fcmp_u32_update_by_TC
+Checking anyof_icmp_u16_update_by_TC
```
fhahn, line 12: I think this needs

```suggestion
Checking anyof_icmp_u16_update_by_TC
exit 0
```

to pass?

fhahn (approved):
LGTM, but needs `exit 0` in the reference output I think. might be good to run and check again to make sure everything passes

Mel-Chen:
Passed with `-mcpu=sifive-x280 -O3 -mllvm -force-tail-folding-style=data-with-evl -mllvm -prefer-predicate-over-epilogue=predicate-dont-vectorize`.

PR #196, opened 2025-01-03
[Fortran/gfortran] Disable bounds-check tests.

vzakhari:
The tests started failing at -O2 after https://github.com/llvm/llvm-project/pull/121544
Flang inlines hlfir.assign with hlfir.expr RHSs after the change,
so the runtime is not called, and it does not make the bounds checks.


tarunprabhu (approved):
Will these tests ever pass in the future? In other words, as far as these tests are concerned, will flang's behavior always be different from gfortran? If so, could you add a comment clarifying this. It will help sort out which tests we can permanently disable.

vzakhari:
Thank you for the review!

These tests will pass with `-O0` as said in the comment.  They will also pass whenever the bounds check feature is implemented.

tarunprabhu:
> These tests will pass with `-O0` as said in the comment. They will also pass whenever the bounds check feature is implemented.

Got it! I'm glad that that comment is still relevant. Thanks for the clarification.

PR #197, opened 2025-01-03
[Fortran/gfortran] Disable tests with unsupported real type on ppc64le and powerpc

kkwli:
The `real(16)` type is not supported on PPC. This patch is to disable those tests on `ppc64le` and `powerpc`.


tarunprabhu (approved):
LGTM. Thanks.

madanial0 (approved):
LGTM!

PR #198, opened 2025-01-06
[Microbench] Sync with google/benchmark to change FindPFM.cmake

DataCorrupted:
libpfm only contains dynamic library in some of our environments. Since we don't actually static link them, only find dynamic library is enough. This also aligns with LLVM's configuration

https://github.com/llvm/llvm-project/blob/7d5376270ae5807a29597a91d7cf59f967ccf39e/llvm/cmake/modules/FindLibpfm.cmake#L13

Cherry-picked from: https://github.com/google/benchmark/commit/a092f8222c1fa95986efe611abf78daba59d3c59

### DIFF
--- a/MicroBenchmarks/libs/benchmark/cmake/Modules/FindPFM.cmake
+++ b/MicroBenchmarks/libs/benchmark/cmake/Modules/FindPFM.cmake
@@ -11,7 +11,7 @@ set_package_properties(PFM PROPERTIES
                        DESCRIPTION "a helper library to develop monitoring tools"
                        PURPOSE "Used to program specific performance monitoring events")
 
-check_library_exists(libpfm.a pfm_initialize "" HAVE_LIBPFM_INITIALIZE)
+check_library_exists(pfm pfm_initialize "" HAVE_LIBPFM_INITIALIZE)
 if(HAVE_LIBPFM_INITIALIZE)
   check_include_file(perfmon/perf_event.h HAVE_PERFMON_PERF_EVENT_H)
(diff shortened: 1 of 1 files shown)


mtrofin:
would this pick the static one if it were available?

DataCorrupted:
> would this pick the static one if it were available?

Looks like it default links dynamic library. More the reason to test for dynamic instead of static library

https://github.com/llvm/llvm-test-suite/blob/a5f02a8aac6204f615da4b1ed59db850c6144ba1/MicroBenchmarks/libs/benchmark/src/CMakeLists.txt#L37

Unfortunately I can't install static library on my end to test it out.

mtrofin:
Since this is a fork off github.com/google/benchmark, which we occasionally re-sync with, can you instead copy the way it's done there?

DataCorrupted:
I updated FindPFM as what benchmark does. But it looks like `benchmark` in llvm-test-suite is relative old and maybe that calls for a re-sync?

DataCorrupted:
> If you want to do a resync, that'd definitely be appreciated!

I just got started with this benchmark, a resync with dozens of file changed is really not my top priority yet. Can we land this first? 

After I have some data from benchmark I'll circle back to the resync.

BTW: Why wasn't benchmark a submodule which would make tracking and syncing easier?

mtrofin:
LGTM, please update the commit message to point to the cherrypicked hash from the `google/benchmark` repo, i.e. [a092f82](https://github.com/google/benchmark/commit/a092f8222c1fa95986efe611abf78daba59d3c59).

Re. submodule - would be worth testing the waters on discourse, see if there are any concerns.

mtrofin (approved):

PR #200, opened 2025-01-13
[SPEC2017] Do not add bwaves_[34].in inputs under "speed" suite type

nmosier:
Do not run 603.bwaves_s on the bwaves_3.in and bwaves_4.in inputs under the reference input size. The speed data dir for 603.bwaves_s does not contain bwaves_3.in and bwaves_4.in, resulting in test failures when running 603.bwaves_s with llvm-lit.


SixWeining:
Included in #247

PR #201, opened 2025-01-15
[Fortran/gfortran] Disable bounds-check tests for matmul tests.

vzakhari:
The tests started failing at -O2 after https://github.com/llvm/llvm-project/pull/122821
Flang inlines hlfir.matmul[_transpose] after the change,
so the runtime is not called, and it does not make the bounds checks.


tarunprabhu (approved):
LGTM. Thanks.

PR #206, opened 2025-02-10
Fix UB in pops.c

fmayer:
It is undefined behaviour to construct a pointer that is out-of-bounds, not just to use it.

Cherry-pick of commit in mimalloc-bench:
https://github.com/daanx/mimalloc-bench/commit/b517ae367c186be7365d238806b475478637d8f2

### DIFF
--- a/MultiSource/Benchmarks/MallocBench/cfrac/pops.c
+++ b/MultiSource/Benchmarks/MallocBench/cfrac/pops.c
@@ -91,10 +91,11 @@ precision palloc(size)
    register posit size;
 {
    register precision w;
-   register cacheType *kludge = pcache + size;	/* for shitty compilers */
+   register cacheType *kludge;  /* for shitty compilers */
 
 #if !(defined(NOMEMOPT) || defined(BWGC))
-   if (size < CACHESIZE && (w = kludge->next) != pUndef) {
+   if (size < CACHESIZE && (kludge = pcache + size) &&
+       (w = kludge->next) != pUndef) {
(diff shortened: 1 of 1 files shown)


MultiSource/Benchmarks/MallocBench/cfrac/pops.c
```diff
@@ -91,10 +91,11 @@ precision palloc(size)
    register posit size;
 {
    register precision w;
-   register cacheType *kludge = pcache + size;	/* for shitty compilers */
+   register cacheType *kludge;  /* for shitty compilers */
```
shafik, line 94: I realize you are just modifying an existing line of code w/ the comment but I don't believe this comment lines up w/ our CoC and so I think a drive by edit is called for here.

fmayer: This is just a cherry-pick of https://github.com/daanx/mimalloc-bench/commit/b517ae367c186be7365d238806b475478637d8f2

vitalybuka: You can put it into description or title

fmayer: I did?

vitalybuka: Thanks, cherry-pick make reader to expect an exact copy.

vitalybuka (approved):

PR #207, opened 2025-02-11
Enable gcc-c-torture suite for hexagon ; Add HVX test for hexagon (#205)

androm3da:
* Add HVX intrinsic test for Hexagon

* Enable gcc torture suite for hexagon


androm3da:
whoops - this PR is bound to the branch name, not the commit.  that won't work, I'll make a new PR or find some other way to land this change on the release/20.x branch.

PR #210, opened 2025-02-22
[Hexagon] Add v79 HVX cmake cache file

androm3da:


androm3da:
@tstellar it looks like the recent 20.1.0 release candidate tags do not tag the latest `release/20.x` commits on `llvm-test-suite`.  Can we update the commit used for `final` / `rc4`?

tstellar:
@androm3da Yes, sorry.   We've never had a push before to the release branch of the test suite, so I didn't have any process to follow for it.  Can you explain why we need these two fixes and what the impact of the changes are?

androm3da:
> @androm3da Yes, sorry. We've never had a push before to the release branch of the test suite, so I didn't have any process to follow for it.

Ah, I see.  Does that mean that we never find bugs during the release that we attribute to the test suite?  Doesn't this kind of thing happen when clang gets stricter and test cases are found to have new warnings/errors?

> Can you explain why we need these two fixes and what the impact of the changes are?

Well they're for improving test coverage for hexagon... so maybe not a *need* so much as a want.  If it can't/shouldn't be included, that's fine and I can workaround it.

PR #212, opened 2025-02-25
[tools] Fix tools/test and not for user-mode emulation

iamlouk:
Add preliminary support for user-mode emulation to the `not` tool
used in some fortran tests by adding a new mode where it spawns
the subprocess under the emulator. The original non-emulator-spawning
version is still needed for tests that use it to verify the output
using the host's `diff`. In the new version, the `TEST_SUITE_RUN_UNDER`
cmake option contents is prepended to the arguments then used to spawn
a subprocess.

Also disables the check_env test in tools/test in case of user-mode
emulation because it would run the host python interpreter in the
target emulator.


iamlouk:
@tarunprabhu Could you maybe have a look at this? Without these changes, some tests in `tools/test` always fails when using `-DTEST_SUITE_USER_MODE_EMULATION=On` and e.g. `-DTEST_SUITE_RUN_UNDER="qemu-<...> -L <...>"`

tools/not.cpp
```diff
@@ -54,6 +58,19 @@ int main(int argc, char* const* argv) {
   if (argc == 0)
     return 1;
 
+#ifdef TEST_SUITE_RUN_UNDER
+  // In case of user-mode emulation, before spawning a new subprocess, the
+  // emulator needs to be preprended to the argv vector for the child.
+  // TEST_SUITE_RUN_UNDER will be defined to a comma-separated list of
+  // string litterals.
```
tarunprabhu, line 65: NIT: 

```suggestion
  // string literals.
```

Fortran/gfortran/CMakeLists.txt
```diff
@@ -716,10 +716,18 @@ function(gfortran_add_execute_test expect_error main others fflags ldflags)
   gfortran_make_working_dir("${target}" working_dir)
   get_filename_component(working_dir_name "${working_dir}" NAME)
 
+  # When running in a emulator, use the version of 'not' that will
```
tarunprabhu, line 719: NIT:

```suggestion
  # When running in an emulator, use the version of 'not' that will
```

Fortran/gfortran/CMakeLists.txt
```diff
@@ -716,10 +716,18 @@ function(gfortran_add_execute_test expect_error main others fflags ldflags)
...
+  # spawn the subprocess under that emulator as well.
+  if(TEST_SUITE_USER_MODE_EMULATION AND TEST_SUITE_RUN_UNDER)
+    set(NOT_HELPER "not-spawning-emulator")
```
tarunprabhu, line 722: Perhaps name this variable `NOT_TOOL`. It's not really a helper.

```suggestion
    set(NOT_TOOL "not-spawning-emulator")
```

tools/test/CMakeLists.txt
```diff
@@ -13,15 +13,26 @@ add_dependencies(ret0 not)
 llvm_test_run(EXECUTABLE "$<TARGET_FILE:not>" "$<TARGET_FILE:not>" "$<TARGET_FILE:ret0>")
 llvm_add_test_for_target(ret0)
 
-# Check that expected crashes are handled correctly.
-llvm_test_executable_no_test(abrt abort.c)
-add_dependencies(abrt not)
-llvm_test_run(EXECUTABLE "$<TARGET_FILE:not>" "--crash" "$<TARGET_FILE:abrt>")
-llvm_add_test_for_target(abrt)
+# These test will always fail under user-mode emulation because 'not'
```
tarunprabhu, line 16: I am not sure what tests you are referring to here. If you mean the `check_env` test that you have removed, then perhaps this comment could be moved after the abort test instead.

tools/test/CMakeLists.txt
```diff
@@ -13,15 +13,26 @@ add_dependencies(ret0 not)
...
+# spawns a subprocess outside the emulator and the check_env test
+# runs the host python interpreter under the emulator for the target.
+if(NOT(TEST_SUITE_USER_MODE_EMULATION AND TEST_SUITE_RUN_UNDER))
```
tarunprabhu, line 19: The negation of the conditional is a bit hard to read. Could the branches be switched with a "positive" conditional instead?

```suggestion
if(TEST_SUITE_USER_MODE_EMULATION AND TEST_SUITE_RUN_UNDER)
```

tools/test/CMakeLists.txt
```diff
@@ -13,15 +13,26 @@ add_dependencies(ret0 not)
...
 llvm_add_test_for_target(ret0)
 
-# Check that expected crashes are handled correctly.
-llvm_test_executable_no_test(abrt abort.c)
-add_dependencies(abrt not)
-llvm_test_run(EXECUTABLE "$<TARGET_FILE:not>" "--crash" "$<TARGET_FILE:abrt>")
-llvm_add_test_for_target(abrt)
+# These test will always fail under user-mode emulation because 'not'
+# spawns a subprocess outside the emulator and the check_env test
+# runs the host python interpreter under the emulator for the target.
```
tarunprabhu, line 18: Does passing environment variables work with `not-spawning-emulator`? Is it only the test here that cannot be run? It would be good to have a test for this if possible.

tarunprabhu (commented):
Do we need a separate executable for this? Could we prepend the arguments when calling `not` instead? Or perhaps pass these on the command line? It seems as if users have to check `TEST_SUITE_USER_MODE_EMULATION` and `TEST_SUITE_RUN_UNDER` and use a different tool anyway, so user code does have to be changed.

tools/not.cpp
iamlouk: Sorry, fixed.

Fortran/gfortran/CMakeLists.txt
iamlouk: Sorry, fixed.

tools/test/CMakeLists.txt
iamlouk: With the `NOT_TOOL` assigned now and the `check_env` test being modified so that it now works under emulation, this is almost back to how it was before (just with `${NOT_TOOL` instead of `not`).

iamlouk:
> Do we need a separate executable for this? Could we prepend the arguments when calling not instead? Or perhaps pass these on the command line?

That was my first idea as well, and I can do that if its preferred. This just seamed easier to me, and other tools like `fpcmp` are also compiled twice (although the context/needs are different as it is compiled for different architectures).

Would a syntax similar to `not [--crash] [--run-under <emulator> <emulator arg1> <emulator arg2 ...> --] <test binary> <test arg1> <test arg2 ...>` be fine?

tools/not.cpp
tarunprabhu: No need to apologize :-)

tarunprabhu:
Thanks for the changes.

> That was my first idea as well, and I can do that if its preferred. This just seamed easier to me, and other tools like `fpcmp` are also compiled twice (although the context/needs are different as it is compiled for different architectures).

I think it might be better to stick with a single tool for this. I am guessing that `fpcmp` needs different CPU and GPU versions. That's not really the case here. This is really just adding a "wrapper" around the invocation of `test_binary`.

> Would a syntax similar to `not [--crash] [--run-under <emulator> <emulator arg1> <emulator arg2 ...> --] <test binary> <test arg1> <test arg2 ...>` be fine?

I assume that `<emulator>  <emulator args> ...` would be in quotes. Otherwise, how would we distinguish between the last `<emulator arg>` and the `<test_binary>`?

iamlouk:
I just pushed a commit with the proposed solution so that you can have a look (single `not` binary, but with a flag to optionally spawn the subcommand under the emulator). The `<emulator>  <emulator args> ...` is *not* in quotes, because that would require spiting arguments in the `not` tool. Instead, a `--` (double dash) is used, a bit like commands like `git checkout` or so sometimes use to disambiguate in similar situations. I just tested this with and without user-mode emulation and with extra flags to the emulator.

tools/test/CMakeLists.txt
```diff
@@ -13,15 +8,21 @@ add_dependencies(ret0 not)
...
-find_package(Python COMPONENTS Interpreter)
+llvm_test_executable_no_test(test_not test_not.cpp)
 llvm_test_executable_no_test(check_env check_env.c)
-add_dependencies(check_env not)
-llvm_test_run(EXECUTABLE ${Python_EXECUTABLE} "%b/test/test_not.py" "$<TARGET_FILE:not>" "$<TARGET_FILE:check_env>")
+add_dependencies(check_env not test_not)
+if(TEST_SUITE_USER_MODE_EMULATION AND TEST_SUITE_RUN_UNDER)
+  # ${TEST_SUITE_RUN_UNDER} is needed here because test_not is not itself
+  # user-mode emulation aware, unlike the not tool.
+  llvm_test_run(EXECUTABLE ${Python_EXECUTABLE} "$<TARGET_FILE:test_not>" ${TEST_SUITE_RUN_UNDER} ${NOT_TOOL} "$<TARGET_FILE:check_env>")
```
tarunprabhu, line 24: Is `${Python_EXECUTABLE}` required here?

CMakeLists.txt
```diff
@@ -319,6 +319,15 @@ if (TEST_SUITE_USER_MODE_EMULATION)
   set(FPCMP fpcmp)
 endif()
 
+# Shortcut for the path to the not executable
+set(NOT_TOOL "$<TARGET_FILE:not>")
+if (TEST_SUITE_USER_MODE_EMULATION AND TEST_SUITE_RUN_UNDER)
+  # Because 'not' spawns a subprocess for its argument, it needs to know about the
```
tarunprabhu, line 325: Nit:

The Github interface doesn't show columns, but does this comment fit within 80 characters? The rules for line width in cmake are not as strict as source files, but it would be good to stick to that width where we can.

tools/not.cpp
```diff
@@ -10,6 +10,12 @@
 //     Will return true if cmd doesn't crash and returns false.
 //   not --crash cmd
 //     Will return true if cmd crashes (e.g. for testing crash reporting).
+//   not --run-under ...<emulator and args>... -- cmd
+//     Will prepend the emulator and its arguments to the spawn of the
+//     subcommand. If --crash is used as well, it has to come after the
+//     '--run-under <...> ...' arguments. The double-dash is used to separate
+//     emulator arguments from cmd arguments. This order makes it easier to
+//     use the not tool in the test-suite.
```
tarunprabhu, line 18: Could you add sentence clarifying what the return value will be in this case.

tarunprabhu (commented):
Thank you for all the changes

tools/test/CMakeLists.txt
iamlouk: Thanks for noticing, not it is not. I just tested in a clean build to make sure.

tools/not.cpp
iamlouk: Just did that. This works as expected with QEMU and another emulator I use downstream, I will double check if this is also the behavior of other user-mode emulators such as Gem5 SE mode or RISC-V Spike.

tools/not.cpp
```diff
@@ -10,13 +10,22 @@
...
 //   not --crash cmd
 //     Will return true if cmd crashes (e.g. for testing crash reporting).
+//   not --run-under ...<emulator and args>... -- cmd
+//     Will prepend the emulator and its arguments to the spawn of the
+//     subcommand. If --crash is used as well, it has to come after the
+//     '--run-under <...> ...' arguments. The double-dash is used to separate
+//     emulator arguments from cmd arguments. This order makes it easier to
+//     use the not tool in the test-suite. The emulator is expected to exit
+//     with the same exit status/signal than the emulated binary in case of
+//     a crash, and the --crash flag will behave the same way.
```
tarunprabhu, line 20: Nit

```suggestion
//     with the same exit status/signal as the emulated binary. The --crash flag 
//     will behave the same way.
```

tarunprabhu (approved):
Thanks for all the changes and for seeing this through.

PR #213, opened 2025-02-25
[flang] Disable erroneous test

klausler:
Disable a test that declares non-SAVE local variables with allocatable coarray components; this is an error that flang-new will shortly be detecting.


tarunprabhu (approved):
If I understand this correctly, this test will (shortly) always raise an error in `flang`  but does not raise an error in gfortran. In that case, could you add a note to the test, something along the lines of:

PR #217, opened 2025-03-05
[flang] Disable two xfail tests after co_reduce support

JDPailleux:
After the merge of https://github.com/llvm/llvm-project/pull/125115 to flang,  two gfortran tests compile and run successfully after the support of `co_reduce` but they're marked "xfail".

### DIFF
--- a/Fortran/gfortran/regression/DisabledFiles.cmake
+++ b/Fortran/gfortran/regression/DisabledFiles.cmake
@@ -1186,6 +1186,8 @@ file(GLOB FAILING_FILES CONFIGURE_DEPENDS
   bind_c_usage_28.f90
   c_loc_tests_11.f03
   coarray_collectives_2.f90
+  coarray_collectives_10.f90
+  coarray_collectives_13.f90
   implicit_14.f90
   pr90290.f90
   pr91564.f90


omjavaid:
@JDPailleux if these tests are passing consistently then shouldnt we remove the xfail decorator instead of disabling them
```
Fortran/gfortran/regression/tests.cmake:compile;coarray_collectives_10.f90;xfail;-fcoarray=single -std=f2008;;
Fortran/gfortran/regression/tests.cmake:compile;coarray_collectives_13.f90;xfail;-fcoarray=single -std=f2008;;

```

JDPailleux:
@omjavaid I don't know if this is the right way to do it because “normally”, when passing the `-std=f2008` flag, this test will fail because `co_reduce` is not supported by the 2008 standard. (like the `coarray_collectives_2.f90` on `co_sum` and more, I've adopted the same way of ignoring it because I think it's better than removing the xfail). 
I think it's best to ignore it for now, but maybe I'm wrong.

omjavaid:
> @omjavaid I don't know if this is the right way to do it because “normally”, when passing the `-std=f2008` flag, this test will fail because `co_reduce` is not supported by the 2008 standard. (like the `coarray_collectives_2.f90` on `co_sum` and more, I've adopted the same way of ignoring it because I think it's better than removing the xfail). I think it's best to ignore it for now, but maybe I'm wrong.

Ack. Agreed Thanks

omjavaid (approved):
LGTM

tarunprabhu (approved):
Thanks.

It seems like theese tests should "xfail" in the future when `flang` implements strict standards checks. In that case, disabling the tests temporarily is the right thing to do. 

There is a mechanism to override the DejaGNU annotations in case we need it. See the README in the `Fortran/gfortran` for details. The source files in the gfortran test suite directory must not be changed.

klausler:
The semantics checks in f18 don't structure features by standard revisions, so don't depend on GNU-like `-fstd=...` options that won't be implemented.

tarunprabhu:
@klausler, 

There are a number of tests in the gfortran test suite that check for standards compliance. Will `flang` ever have something similar?

PR #220, opened 2025-03-13
[Fortran] Add floating point to integer conversion test

ashermancinelli:
Semantics for floating point conversions to integer types was changed in flang to use the saturated floating point intrinsics to more closely match the standard semantics in [llvm-project#130686](https://github.com/llvm/llvm-project/pull/130686) . This tests conversions from tiny, huge, and inf to various integer types.


jeanPerier (approved):
Thanks for adding this test here! Works for me.

tarunprabhu:
>     * We can remove unsigned if folks use this test suite with compilers that don't support that flag (or disable this test under some flag)

I don't think the `llvm-test-suite` is really intended to be a general-purpose compiler test suite. Unless `flang` has a way to disable support for `unsigned`, I wouldn't worry about it.

>     * I wouldn't mind adding real(kind=16) support if we have a way to tell if it's supported with the current compiler

I was thinking that myself but I don't know how to check for this off the top of my head. Maybe @jeanPerier knows, or knows of someone who does.

ashermancinelli:
Thanks @tblah! If this test suite is not trying to be portable, maybe we could use the `FLANG_SUPPORT_R16` macro?

tarunprabhu (approved):
This worked for me on Linux on x86. Thanks.

ashermancinelli:
I split the test into two, one of which uses real(16), and the latter is conditionally enabled. 

Also, since this is just for LLVM I added a quick check for NaN -> int conversions.

Thanks for the reviews! Please LMK if there's anything else y'all would like changed.

tblah (approved):
Thanks for the update.

This correctly enabled r16 support and passed on my aarch64 system.

tarunprabhu (commented):
Does `fp_convert_r16.f90` need to do the same thing as `fp_convert.f90` in addition to `kind=16`? It looks like `fp_convert.f90` is always built and run. Could we do just `kind=16` in the other file then or am I missing something obvious>

ashermancinelli:
> Does fp_convert_r16.f90 need to do the same thing as fp_convert.f90 in addition to kind=16?

The fp_convert_r16 test converts to some of the same types, but from real16 and also to int16 and unsigned16 - there should be no conversions that happen in both test cases. Is that what you were asking?

tarunprabhu (approved):
> > Does fp_convert_r16.f90 need to do the same thing as fp_convert.f90 in addition to kind=16?
> 
> The fp_convert_r16 test converts to some of the same types, but from real16 and also to int16 and unsigned16 - there should be no conversions that happen in both test cases. Is that what you were asking?

Ah, sorry. I didn't notice that value in `do_conversion` is `real(8)` in one and `real(16)` in the other. 

Thanks for clarifying. LGTM

PR #221, opened 2025-03-13
[utils] Fix Warning: Series.__getitem__ treating keys as positions  is deprecated

sunshaoce:
warning log:
```
./utils/compare.py:206: FutureWarning: Series.__getitem__ treating keys as positions is deprecated. In a future version, integer keys will always be treated as labels (consistent with DataFrame behavior). To access a value by position, use `ser.iloc[pos]`
  name0 = names[0]
```


lukel97 (approved):
Oh nice, that warning's been bugging me for a while. Thanks!

PR #223, opened 2025-03-20
[Prolangs-C] Fix compilation errors with clang 15 and above

androm3da:
Correcting these function declarations addresses these compilation errors:

```
llvm-test-suite/MultiSource/Benchmarks/Prolangs-C/loader/sym_tab.c:40:5: error: conflicting types for 'INSERT_IN_SYM_TAB'
   40 | int INSERT_IN_SYM_TAB(char *MODULE,char *LABEL,int LOCATION,enum kind TYPE,
      |     ^
llvm-test-suite/MultiSource/Benchmarks/Prolangs-C/loader/sym_tab.h:52:12: note: previous declaration is here
   52 | extern int INSERT_IN_SYM_TAB();
      |            ^
1 error generated.

llvm-test-suite/MultiSource/Benchmarks/Prolangs-C/assembler/sym_tab.c:40:5: error: conflicting types for 'INSERT_IN_SYM_TAB'
   40 | int INSERT_IN_SYM_TAB(char *MODULE,char *LABEL,int LOCATION,enum kind TYPE,
      |     ^
llvm-test-suite/MultiSource/Benchmarks/Prolangs-C/assembler/sym_tab.h:52:12: note: previous declaration is here
   52 | extern int INSERT_IN_SYM_TAB();
      |            ^
1 error generated.
```

Since clang 15.x, this has been an error for C programs.

### DIFF
--- a/MultiSource/Benchmarks/Prolangs-C/assembler/sym_tab.h
+++ b/MultiSource/Benchmarks/Prolangs-C/assembler/sym_tab.h
@@ -49,6 +49,6 @@ extern struct SYMBOL_TABLE_ENTRY *LOOK_UP_SYMBOL();
 /* If <MODULE,LABEL> pair is not in the symbol table, put it there and       */
 /* TRUE (integer 1). Otherwise return FALSE (integer 0).                     */
 /* Puts onto front of the linked list.                                       */
-extern int INSERT_IN_SYM_TAB();
+extern int INSERT_IN_SYM_TAB(char *, char *, int, enum kind, SYMBOL_TABLE *);
 
 
--- a/MultiSource/Benchmarks/Prolangs-C/loader/sym_tab.h
+++ b/MultiSource/Benchmarks/Prolangs-C/loader/sym_tab.h
@@ -49,6 +49,6 @@ extern struct SYMBOL_TABLE_ENTRY *LOOK_UP_SYMBOL();
 /* If <MODULE,LABEL> pair is not in the symbol table, put it there and       */
 /* TRUE (integer 1). Otherwise return FALSE (integer 0).                     */
 /* Puts onto front of the linked list.                                       */
-extern int INSERT_IN_SYM_TAB();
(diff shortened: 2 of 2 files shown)


DavidSpickett:
I think they would have been downloaded from this page - https://prolangs.cs.vt.edu/rutgers/software.php.

This is a collection of existing software, put together some time around 1997 going by the modification dates. Used for some research, so 1. they wouldn't change it because it needs to match that research and 2. all these would have their own upstream locations, assuming they still exist. This one I've no idea how I'd even find it.

We have code changes made to it already: https://github.com/llvm/llvm-test-suite/commits/main/MultiSource/Benchmarks/Prolangs-C/assembler

It's not under a specific license but includes:
```
/* %%%%%%%%%%%%%%%%%%%% (c) William Landi 1991 %%%%%%%%%%%%%%%%%%%%%%%%%%%% */
/* Permission to use this code is granted as long as the copyright */
/* notice remains in place. */
```
So for some definition of "use" we are still complying with that.

TLDR: it's fine to modify this here in llvm-test-suite, or at least, as fine as it was for the previous changes :)

Also surprised this hasn't come up before.

Would be good to include in the commit message why this error happens and what version of clang started doing it.

DavidSpickett:
I think this is it:
https://releases.llvm.org/15.0.0/tools/clang/docs/ReleaseNotes.html#improvements-to-clang-s-diagnostics

> Clang now appropriately issues an error in C when a definition of a function without a prototype and with no arguments is an invalid redeclaration of a function with a prototype. e.g., void f(int); void f() {} is now properly diagnosed.

https://godbolt.org/z/7q8zbPTcq

15 is the first one to error here.

androm3da:
> Would be good to include in the commit message why this error happens and what version of clang started doing it.

Thanks - I have updated the commit message.

DavidSpickett:
I should have been clearer, the commit message is now fine, but for it to be used for the final commit message when we "Squash and merge" it needs to be placed in the PR's description as well.

DavidSpickett:
Even if you are the first to hit this (which I doubt), I can't imagine you would have been the last. So thanks for contributing a fix!

PR #229, opened 2025-04-04
Remove pre-compiled binaries

nikic:
For some reason, we have three pre-compiled binaries checked in for the gs, agrep and football benchmarks. These are not actually used (we'll build our own, as usual). Presumably they were committed by accident.
    
We also ship one more pre-compiled makeproto binary for the p2c benchmark. This is not necessary either, as we already ship the `.proto` files it generates.


DavidSpickett:
At least was here since the project was imported - https://github.com/llvm/llvm-test-suite/commit/a23d3270f9c36b4a1f6f54133c4374565e555abd.

nikic:
> I assume you used some command to search to see if there were any others?

Nope, these were just flagged by some compliance tooling. But we remove some subdirectories for licensing reasons, so there are actually more... I did a more thorough check with `find . -type d -name 'build*' -prune -o -type f -exec file {} + | grep "ELF.*executable" | cut -d: -f1` and dropped the genarch and makeproto binaries as well.

Worth noting that makeproto is different from the rest in that we don't seem to ship the source for it, but we do ship the `.proto` files it generates, so having the binary shouldn't be necessary either.

PR #232, opened 2025-04-15
[Fortran] disable pointer_check_11.f90 which exhibits UB

tblah:
This test calls an empty subroutine with a null pointer as an argument. In gfortran it is hoped that a flag will detect this and produce an error. Flang does not perform any such analysis, and so on some machines when optimization is enabled, this test never terminates. I don't think this is a flang bug as such because the input code is doing undefined behavior.

### DIFF
--- a/Fortran/gfortran/regression/DisabledFiles.cmake
+++ b/Fortran/gfortran/regression/DisabledFiles.cmake
@@ -1046,7 +1046,7 @@ file(GLOB FAILING_FILES CONFIGURE_DEPENDS
   namelist_96.f90 # real data for integer NAMELIST input
   no_unit_error_1.f90
   pointer_check_10.f90
-  pointer_check_11.f90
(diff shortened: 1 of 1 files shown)


pawosm-arm:
I can confirm this is happening. Build with -O0 (which is the default for the Debug build of this test suite), this program crashes as expected. Built with -O3 (which is the default for the Release build) it spins forever until is hit by the 2h timeout (and this is reported as passing successfully after all!). Since this is UB, all bets are off, and compiler is allowed to do whatever when optimizing it at -O3, so one can't claim it as incorrect's compiler behavior.

pawosm-arm (approved):

tarunprabhu (approved):
Could you add a comment describing why it is being disabled? At some point, the disabled tests need to be reviewed and a comment like this would be a great help. 

Thanks.

pawosm-arm:
Could you consider cherry-picking it to the release/20.x branch?

PR #235, opened 2025-04-18
[Fortran/gfortran] Disable bounds-check tests for maxloc.

vzakhari:
The tests will start failing at -O2 when I merge
https://github.com/llvm/llvm-project/pull/136071
The HLFIR inlining does not perform the bounds check.


tarunprabhu (approved):
Thanks.

PR #236, opened 2025-04-21
[SPEC] Workaround error with latest clang in 523.xalancbmk_r

lukel97:
523.xalancbmk_r has some code that calculates the offset of a member, but is undefined behaviour.

After https://github.com/llvm/llvm-project/pull/130742 this is now optimized away which results in a miscompile.

This has been fixed in upstream versions of xerces-c, but not yet in SPEC CPU 2017: https://github.com/apache/xerces-c/commit/02e48494496dd24476490fd36c1bc97b6a37002e#diff-e2f4677367dcf43bd6d31b2bbca5b1aa89e36ec155a040fe4447b201a554cf81

This patch passes -fwrapv-pointer to 523.xalancbmk_r to work around it for now. Ideally SPEC CPU 2017 will issue an update which fixes this issue and we can drop this flag.

I've included it in the 2006 benchmark too for good measure, although I haven't tried it out, see 92b58b3bf4c6001bff4c7c7f8f98305e3ead56ff

### DIFF
--- a/External/SPEC/CINT2006/483.xalancbmk/CMakeLists.txt
+++ b/External/SPEC/CINT2006/483.xalancbmk/CMakeLists.txt
@@ -14,6 +14,11 @@ add_definitions(
 
 list(APPEND CXXFLAGS -std=gnu++98)
 
+# Workaround undefined behaviour in xerces-c dependency.
+# This is fixed upstream but not included in SPEC CPU 2017:
+# https://github.com/apache/xerces-c/commit/02e48494496dd24476490fd36c1bc97b6a37002e#diff-e2f4677367dcf43bd6d31b2bbca5b1aa89e36ec155a040fe4447b201a554cf81
+add_compile_options(-fwrapv-pointer)
+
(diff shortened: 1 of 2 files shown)


madhur13490:
+1. this is needed not just for SPEC2006 but also for the same benchmark in SPEC 2017 as per our CI failure.

mikhailramalho (approved):
That fixed 523.xalancbmk_r for me, thanks!

lukel97:
> Does this have any notable impact on codegen, i.e. will it cause perf differences?

It's likely going to have a negative impact, I'm not sure how much though. The RISC-V LNT runners should finish a run by tomorrow and we should have an idea. 

I'm open to any better alternative options by the way. It's unfortunate that we don't have control over the SPEC benchmarking code.

lukel97:
@citymarina good point, I've opened https://github.com/llvm/llvm-test-suite/pull/237 for this. Does that fix any issues you're seeing with older clangs?

PR #237, opened 2025-04-22
[SPEC] Add compiler check for -fwrapv-pointer in xalancbmk

lukel97:
-fwrapv-pointer is a relatively recent flag and older versions of clang may not support it.

Hopefully older versions of clang also don't perform the UB optimisation that warrants the use of the flag in the first place

### DIFF
--- a/External/SPEC/CINT2006/483.xalancbmk/CMakeLists.txt
+++ b/External/SPEC/CINT2006/483.xalancbmk/CMakeLists.txt
@@ -14,12 +14,17 @@ add_definitions(
 
 list(APPEND CXXFLAGS -std=gnu++98)
 
+include(CheckCXXCompilerFlag)
+
 # Workaround undefined behaviour in xerces-c dependency.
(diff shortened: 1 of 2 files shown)


citymarina:
I'm not currently in front of the right machine to test, but would `add_cxx_compiler_flag` take care of this check for you?

lukel97:
> I'm not currently in front of the right machine to test, but would `add_cxx_compiler_flag` take care of this check for you?

It looks like that's only available in the MicroBenchmarks directory, I can't seem to import `include(AddCXXCompilerFlag)` from the xalancbmk module.

Maybe it would be worthwhile extracting that out in a separate PR so we can use it in other places? It looks useful

citymarina:
> It looks like that's only available in the MicroBenchmarks directory, I can't seem to import `include(AddCXXCompilerFlag)` from the xalancbmk module.

Oops, I missed that. LGTM then.

PR #239, opened 2025-04-28
[Fortran] Disable gfortran volatile8 test

ashermancinelli:
This test expects the compiler to give a compilation error. Current LLVM flang does emit an error, but it's a TODO message because volatile variables are not handled yet. There are upcoming PRs to enable this though, and the front end ought to emit some diagnostic about this. Eugene E and Peter K are discussing whether these should be warnings or errors. While this is being discussed, disable the test so work on volatile can proceed.

[LLVM PR requiring this change](https://github.com/llvm/llvm-project/pull/132486)
[LLVM issue discussing how to handle this test case](https://github.com/llvm/llvm-project/issues/137369#issuecomment-2831344238)

CC @kiranchandramohan @eugeneepshteyn @vzakhari

### DIFF
--- a/Fortran/gfortran/regression/DisabledFiles.cmake
+++ b/Fortran/gfortran/regression/DisabledFiles.cmake
@@ -1053,6 +1053,8 @@ file(GLOB FAILING_FILES CONFIGURE_DEPENDS
   binding_label_tests_26b.f90
   test_common_binding_labels_2_main.f03
   string_1.f90 # Expect error on 32 bits platform
+  volatile8.f90 # Expects error on bad dummy argument declarations
+                # llvm-project#137369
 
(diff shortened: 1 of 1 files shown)


tarunprabhu (approved):
Once the discussion on how to handle this is resolved, could you update the comment to provide a bit more detail? At some point, we need to go through and decide which tests should be permanently disabled, and which could be handled differently*. A more detailed comment would be of great help. Thanks!

\* There is a mechanism to override the 'xfail' status of a test without modifying the test itself).

ashermancinelli:
Yes! We intend on re-enabling this test, so it should not be disabled for long. Just so buildbots are green while we're in transition. Thank you!

PR #243, opened 2025-05-13
[HIP] rough way of adding cuda_wrapper tests from Externals/CUDA

jmmartinez:
This PR tries to adapt in a pretty rough way some tests from `Externals/CUDA` that test interactions with `cuda_wrapper` headers.

This catch the bug fixed by https://github.com/llvm/llvm-project/pull/139697


jmmartinez:
I'm not happy with this. Ideally, we could put CUDA and HIP common tests together. This is a first step to get some initial feedback (and see what the validation has to say about it).

jmmartinez:
@jplehr do you have some advice about how I can ensure that this change doesn't break any of our buildbots ?

I've run the test locally and it works. I still have to ensure that `hipify-perl` is installed on our test images.

jplehr:
I believe we currently do not install this as part of the CI images, so we need to update the container images that we run. Let me check and get back to you.

jmmartinez:
@jplehr I double checked and this PR is already not failing if hipify-perl is not found.

It shows:

```
CMake Warning at External/HIP/CMakeLists.txt:117 (message):
  hipify-perl not found for ROCm installation in /Externals/hip/rocm-6.3.0.
  External/HIP/CMakeLists.txt:141 (create_local_hip_tests)                                                                                                                                                         External/HIP/CMakeLists.txt:196 (create_hip_test)
  External/HIP/CMakeLists.txt:267 (create_hip_tests)
```

If there are no extra remarks we could:
1. land this (it should not break the production bots since it tolerates hipify not being installed)
2. check what staging bots are telling us and if everything is ok
3. move the staging bots into production

Maybe on Tuesday, cause I don't want to break everything before a 3 day weekend.

What do you all think ?

jplehr:
Our bots should pick up changes to llvm-test-suite whenever we run a build of LLVM.
So, I don't mind landing this now (or on Tuesday) and see what it does to our bots. We can revert if it makes the bot unhappy. Worst case: The revert is not picked up and we need to move one bot back to staging, which isn't the end of the world either.

Whatever works better for you.

jplehr (approved):
LGTM

jmmartinez:
I'll give it a go today then. I'll watch the bots:

[staging bot](https://lab.llvm.org/staging/#/builders/93)
[production bot](https://lab.llvm.org/staging/#/builders/123)

PR #244, opened 2025-05-13
[CUDA][HIP] add test for uses of std::array on device code

jmmartinez:
First attempt at testing https://github.com/llvm/llvm-project/pull/136133 .

Draft for the moment, I have to manually test this further.


External/CUDA/array.cu
```diff
@@ -0,0 +1,71 @@
...
+#if __cplusplus >= 201402L && STDLIB_VERSION >= 2014
+  constexpr size_t N = test_array();
+  (void)(N);
+#endif
+}
+
+__global__ void kernel() { do_test(); }
+
+int main() {
+  kernel<<<32, 32>>>();
```
yxsamliu, line 51: can std::array be parameter type of kernel or device functions?

can it be used as non-constexpr stack variable in kernel or device functions?

if so, can we have testcases for those?

External/HIP/cuda2hip.h
```diff
@@ -0,0 +1,10 @@
+#ifndef CUDA_2_HIP
+#define CUDA_2_HIP
+
+#include "hip/hip_runtime.h"
+
+#define cudaError_t hipError_t
+#define cudaSuccess hipSuccess
+#define cudaDeviceSynchronize hipDeviceSynchronize
```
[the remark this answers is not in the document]
jmmartinez: I've tried with `hipify-perl` and it works.

`hipify-clang` did not, it yield errors for every test.

One common error case was with multiple main definitions (although they are on different ifdef branches):

```cpp
#include <stdio.h>
#if __cplusplus < 201103L
int main() { puts("foo\n"); return 0; }
#else
int main() { puts("bar\n"); return 0; }
#endif
```

```
/tmp/foo.cu-5f2589.hip:5:5: error: redefinition of 'main'
    5 | int main() { puts("bar\n"); return 0; }
      |     ^
/tmp/foo.cu-5f2589.hip:3:5: note: previous definition is here
    3 | int main() { puts("foo\n"); return 0; }
```

I'll adapt these tests to run with `hipify` and remove the hacky header.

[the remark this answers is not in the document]
jmmartinez: I took care of this remark on https://github.com/llvm/llvm-test-suite/pull/243. I'll rebase this PR on top of it.

External/CUDA/array.cu
jmmartinez: > can std::array be parameter type of kernel or device functions?

It can be used as a parameter for device and kernel functions (which would trigger the copy-constructor). According to the c++ reference `std::array` is an aggregate in the the same way as a struct with an array inside; so the same rules should apply. For example, all the constructors are implicitly declared.

I'm not sure that adding the copy-constructor case here will improve the coverage of the test (we're testing the interaction with the standard C++ library, and there is no code to test since these behaviors are implicitly defined by clang). Passing an aggregate as an argument should be covered by the tests on clang's side.

I can add the test for `std::array` constructors here too if you think we really need them.

> can it be used as non-constexpr stack variable in kernel or device functions?

This is already covered in the test. We check `std::array` both in a constexpr and non-constexpr context.

jmmartinez:
Would you be ok to:
1. land this change and see how the build-bots behave.
2. then land https://github.com/llvm/llvm-project/pull/136133

Artem-B:
The commit appears to break CUDA tests:
https://lab.llvm.org/buildbot/#/builders/69/builds/22366  (see the "Configuring CUDA test-suite" section logs)

```
CMake Error at cmake/modules/TestSuite.cmake:48 (add_executable):
  No SOURCES given to target: array-cuda-12.4-c++20-libstdc++-9
  cmake/modules/TestSuite.cmake:79 (llvm_test_executable_no_test)
  cmake/modules/GPUTestVariant.cmake:48 (llvm_test_executable)
  cmake/modules/GPUTestVariant.cmake:71 (create_one_local_test_f)
  External/CUDA/CMakeLists.txt:77 (create_one_local_test)
  External/CUDA/CMakeLists.txt:170 (create_local_cuda_tests)
  External/CUDA/CMakeLists.txt:347 (create_cuda_test_variant)
  External/CUDA/CMakeLists.txt:376 (create_cuda_tests)
```

External/CUDA/CMakeLists.txt
```diff
@@ -63,6 +63,7 @@ macro(create_local_cuda_tests VariantSuffix)
   list(APPEND CUDA_LOCAL_TESTS assert)
   list(APPEND CUDA_LOCAL_TESTS axpy)
   list(APPEND CUDA_LOCAL_TESTS algorithm)
+  list(APPEND CUDA_LOCAL_TESTS array)
```
Artem-B, line 66: There's no matching array.cu file committed under CUDA which makes cmake configuration fail.

PR #246, opened 2025-05-14
[Fortran/gfortran] Skip some directories when generating test configurations

tarunprabhu:
Some directories in the gfortran test suite only contain "dependent" sources. These are sources used in multi-file tests where only the "main" source file contains DejaGNU directives that are used to determine the behavior of the test. A list of such "dependents-only" directories have been added to the update-test-config script so a test configuration is not generated for them.

### DIFF
--- a/Fortran/gfortran/utils/update-test-config.py
+++ b/Fortran/gfortran/utils/update-test-config.py
@@ -163,6 +163,20 @@ def get_lines(filepath: str) -> list[str]:
     finally:
         return lines
 
+# Check if a test configuration should be generated for the given directory
+#
+# Multi-file tests consist of a single "main" file and some number of dependent
+# sources. Some directories only contain "dependent" sources. A test
+# configuration should not be generated for such directories.
+def should_generate_config(d: str) -> bool:
+    # The directories containing only dependent sources. These should be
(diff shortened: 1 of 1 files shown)


Fortran/gfortran/utils/update-test-config.py
```diff
@@ -163,6 +163,20 @@ def get_lines(filepath: str) -> list[str]:
...
 
+# Check if a test configuration should be generated for the given directory
+#
+# Multi-file tests consist of a single "main" file and some number of dependent
+# sources. Some directories only contain "dependent" sources. A test
+# configuration should not be generated for such directories.
+def should_generate_config(d: str) -> bool:
+    # The directories containing only dependent sources. These should be
+    # relative to the root directory containing the gfortran tests.
+    skip: list[str] = ['regression/coarray/add_sources']
```
tblah, line 174: This won't work on Windows (which separates names with `\`). I think instead of a string you could represent with a type intended to handle OS differences in paths? I'm not that familiar with Python but some searching found https://docs.python.org/3/library/pathlib.html#module-pathlib

tblah (commented):
Thanks

tblah (approved):
Thanks for the update

PR #248, opened 2025-05-20
[Fortran/gfortran] Disable little endian test on big endian target

kkwli:


tarunprabhu (approved):
LGTM. Thanks!

Fortran/gfortran/regression/tests.cmake
```diff
@@ -6561,7 +6561,7 @@ run;unsigned_18.f90;;-funsigned;;
 run;unsigned_19.f90;;-funsigned;;
 run;unsigned_2.f90;;-funsigned;;
 run;unsigned_20.f90;;-funsigned;;
-run;unsigned_21.f90;;-funsigned;;
+run;unsigned_21.f90;;-funsigned;;powerpc-.+-aix
```
DanielCChen, line 6564: What is the difference between a single `;` and double `;;`? 
It seems some similar cases use a single `;` for the target.

DanielCChen: I see it now. It skips the <enabled-on> part.

kkwli: @DanielCChen 
This file is generated by `update-test-config.py`. Each line in `test.cmake` consists of semicolon-separated fields.
```
<kind>;<sources>;<xfail>;<options>;<enabled-on>;<disabled-on>
```
`;;` means the field is empty.

Ref: https://github.com/llvm/llvm-test-suite/tree/main/Fortran/gfortran#notes-for-developersmaintainers

PR #249, opened 2025-05-20
[Fortran/gfortran] Disable failing test

tarunprabhu:
This test should fail to compile with an "invalid character in name" error - or the equivalent in flang. However, it used to fail because the LINEAR clause was not implemented in the OpenMP DO construct. The clause is now supported, but flang does not issue the expected compile-time error which causes the test to fail.


kiranchandramohan (approved):
LG.

tblah (approved):

PR #250, opened 2025-05-20
[Fortran/gfortran] Better handling of unsupported gfortran options

tarunprabhu:
A number of tests use gfortran-specific options which are either not supported in flang, or for which a suitable equivalent has not yet been enabled. These options used to be filtered out in CMakeLists.txt when configuring the test suite. This has been moved to the test configuration script to allow for more flexibility (since cmake is a terrible language in which to do anything even remotely sophisticated).

Doing this has allowed us to perform wildcard matches for options that expose gfortran's internal structure which is obviously not relevant for flang. In the future, we do not have to constantly update the filter with these options. We can also distinguish between the option name and its value, so we can filter simply based on the option spelling.

The test configuration files have been regenerated after this change. Some tests fail with this updated configuration and they have been disabled.

------------------
Notes for reviewers:

- Only the following files need to be reviewed. The remaining files have been generated by `update-test-config.py`:
  - [Fortran/gfortran/utils/update-test-config.py](https://github.com/llvm/llvm-test-suite/pull/250/files#diff-8df8668b5da1b6a219777714011fbc814636344dbe43c9cc653da4296e46bee9)
  - [Fortran/gfortran/CMakeLists.txt](https://github.com/llvm/llvm-test-suite/pull/250/files#diff-f85fabb8751d614f90ebd72b4d39a3c90b40f092b9c55ff20c774918d5565124)
  - [Fortran/gfortran/regression/DisabledFiles.cmake](https://github.com/llvm/llvm-test-suite/pull/250/files#diff-eb0bee91f1aa9ac617939b2e138cb659da256de8675932e3ed73647dd3cff26f)
- As a result of this change, several more tests, which are currently disabled, can be enabled. That will be done in a separate PR.
- While I do not anticipate any discrepancies between systems, I request reviewers to check that the test suite passes on their respective platforms.


tblah (approved):
Changes look good to me and this passes on aarch64 for me

jeanPerier (approved):
This works for me on x86-64, thanks!

kkwli (approved):
Thanks

ylzsx (approved):
It works well for me on LoongArch, thanks!

PR #251, opened 2025-05-29
fpcmp: Allow decimal numbers ending with a period

SixWeining:
After 8a84a507d7, fpcmp requires at least one post-decimal digit when period present. This leads to incorrect comparison results for 628.pop2_s of SPEC2017 because this program outputs `65.` if the Fortran compiler is flang while the reference output is `65.0`.


SixWeining:
> Thanks for working on this. Could you add one or more tests in [tools/test](https://github.com/llvm/llvm-test-suite/tree/main/tools)? There are some there that you may be model off of, if needed. That may help prevent regressions like this.

Thanks. Added.

tools/test/CMakeLists.txt
```diff
@@ -19,6 +23,10 @@ add_dependencies(abrt not)
 llvm_test_run(EXECUTABLE "$<TARGET_FILE:not>" "--crash" "$<TARGET_FILE:abrt>")
 llvm_add_test_for_target(abrt)
 
+# Check that fpcmp can handle decimal numbers ending with a period correctly.
+llvm_test_run(EXECUTABLE "$<TARGET_FILE:fpcmp-target>" "-a" "0.03" "-r" "0.03" "-i" "%b/test/fpcmp-input1" "%b/test/fpcmp-input2")
+llvm_add_test_for_target(fpcmp-target)
+
```
tarunprabhu, line 29: Could you move these to the end of the file? The lines of code on either side of this block are related to testing the `not` utility.

tools/test/fpcmp-input1
```diff
@@ -0,0 +1,2 @@
+65.
```
tarunprabhu, line 1: Can we also check for the case where there is no decimal point?

tools/test/CMakeLists.txt
```diff
@@ -19,6 +23,10 @@ add_dependencies(abrt not)
 llvm_test_run(EXECUTABLE "$<TARGET_FILE:not>" "--crash" "$<TARGET_FILE:abrt>")
 llvm_add_test_for_target(abrt)
 
+# Check that fpcmp can handle decimal numbers ending with a period correctly.
+llvm_test_run(EXECUTABLE "$<TARGET_FILE:fpcmp-target>" "-a" "0.03" "-r" "0.03" "-i" "%b/test/fpcmp-input1" "%b/test/fpcmp-input2")
```
tarunprabhu, line 27: The tolerance of 0.03 here seems fairly large. Do you expect that it will fail if it is made smaller?

tarunprabhu (changes requested):
Thanks for adding the tests.

tools/test/CMakeLists.txt
tarunprabhu: > 628_pop2 uses 0.03 for both absolute and relative tolerance.
> 
> ```
> # SPEC Test has both absolute and relative tolerance with same order
> speccpu2017_verify_output(IGNORE_WHITESPACE ABSOLUTE_TOLERANCE 0.03 RELATIVE_TOLERANCE 0.03)
> ```
> 
> How about using a bigger value in the test file? For example: change `3.0E-02` to `1003.0E-02`.

Thanks for the clarification. Since this was motivated by 628_pop2 anyway, the tolerance is probably appropriate. There's no need for larger values.

tarunprabhu (approved):
Thanks for the changes. We should consider adding more tests to explicitly check for error conditions, whitespace etc. but that can be done separately. 

LGTM.

PR #252, opened 2025-05-29
[gfortran] Fixed cray_pointers_2.f90 aliasing issues.

vzakhari:
The test started failing after https://github.com/llvm/llvm-project/pull/140803
The issue is that the Cray pointee and the underlying storage
are both accessed in subroutines violating Fortran aliasing rules
for the dummy arguments.

GCC tried to "solve" that by using `-fno-inline` option, but
it should be better to make the aliasing explicit with proper `TARGET`
attribute placement.

### DIFF
--- a/Fortran/UnitTests/CMakeLists.txt
+++ b/Fortran/UnitTests/CMakeLists.txt
@@ -1,6 +1,7 @@
 # This file should only contain add_subdirectory(...) one for each test
 add_subdirectory(hello)
 add_subdirectory(assign-goto)
+add_subdirectory(cray_pointers_2)
 add_subdirectory(execute_command_line)
 add_subdirectory(fcvs21_f95) # NIST Fortran Compiler Validation Suite
 add_subdirectory(finalization)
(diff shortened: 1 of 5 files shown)


tarunprabhu:
Thanks Slava.

We don't allow modifying these tests. For now, this test should be disabled instead by adding it to the "skipped" tests in `Fortran/gfortran/regression/DisabledFiles.cmake`. Please add a comment explaining why this was disabled (the contents of your commit message would work well).

If this test is useful, you could consider adding it with your modifications to `Fortran/UnitTests`. You can use the [assign-goto](https://github.com/llvm/llvm-test-suite/tree/main/Fortran/UnitTests/assign-goto) test as a template. I believe that that test is a modified version of one from the gfortran test suite.

tarunprabhu (changes requested):
Please disable the test instead.

ashermancinelli:
Thanks for the explanation @tarunprabhu!

As you said Slava, this test appears to break the rules for cray pointers that the HPE docs lay out. [Our own docs on aliasing](https://flang.llvm.org/docs/Aliasing.html#cray-pointers) explain this to our users and how they can get around the issues that come from their misuse of the feature. It may be a useful standalone test demonstrating that the workaround we recommend to our users really does allow their code to work as they intend. I don't like gfortran's approach of simply disabling some optimizations to get the test to pass. +1 for disabling the test and adding our own modified test.

vzakhari:
Thank you for the review, Tarun!

PR #257, opened 2025-06-16
[CUDA] Remove 'array' from CUDA tests as it's only present in HIP

Artem-B:
Unbreaks CUDA test bots broken by #244


jmmartinez:
Argh ! Sorry, I missed that change. I thought I've reverted them.

PR #258, opened 2025-06-17
[HashRecognize] Add tests to verify CRC optz

artagnon:
In preparation to optimize cyclic-redundancy-check loops using table-lookup using the newly introduced HashRecognize analysis, add tests with CRC loops (which will get optimized), and check the result against a Sarwate table-lookup (which will not get optimized). The companion change is llvm/llvm-project#143208.


SingleSource/UnitTests/HashRecognize/crc16.be.c
```diff
@@ -0,0 +1,50 @@
...
+  for (unsigned int i = 1; i < 256; i <<= 1) {
+    crc = (crc << 1) ^ (crc & 0x8000 ? GENPOLY : 0);
+    for (unsigned int j = 0; j < i; j++)
+      CRCTable[i + j] = crc ^ CRCTable[j];
+  }
+}
+
+static uint16_t crc_table(uint16_t crc_initval, uint16_t data) {
+  uint16_t crc = crc_initval;
+  crc_init();
```
pfusik: Call `crc_init` once.

SingleSource/UnitTests/HashRecognize/crc16.be.c
```diff
@@ -0,0 +1,50 @@
...
+    uint16_t crc_shl = crc << 1;
+    crc = (xor_crc_data & 0x8000) ? (crc_shl ^ GENPOLY) : crc_shl;
+    data <<= 1;
+  }
+  return crc;
+}
+
+int main() {
+  uint16_t crc_initval[8] = {0, 129, 11, 255, 16, 4129, 16384, 1};
+  uint16_t data[8] = {0, 1, 11, 16, 129, 255, 4129, 16384};
```
pfusik:
```suggestion
  static const uint16_t crc_initval[8] = {0, 129, 11, 255, 16, 4129, 16384, 1};
  static const uint16_t data[8] = {0, 1, 11, 16, 129, 255, 4129, 16384};
```

SingleSource/UnitTests/HashRecognize/crc16.be.c
```diff
@@ -0,0 +1,50 @@
+#include <stdint.h>
+#include <stdlib.h>
+
+#define GENPOLY 4129
+
+uint16_t CRCTable[256];
```
pfusik:
```suggestion
static uint16_t CRCTable[256];
```

SingleSource/UnitTests/HashRecognize/crc16.be.c
pfusik: Why these particular 8 values?

artagnon: Not sure what to do about the hand-picked values: we'd be over-engineering a bit if we go through all values from INT_MIN to INT_MAX, so I've selected some random values.

pfusik: Hand-picked values are fine, but having 8 of these made me wonder if there's a formula for them. I suggest adding some more, or a comment that these are just some arbitrary test cases.

pfusik (approved):
LGTM

PR #259, opened 2025-06-18
Partially fix test failures with libstdc++ 15 in debug mode

T0b1-iOS:
When compiling on Fedora 42, there are now test failures in debug mode (c.f. llvm/llvm-project#144678).
This PR fixes the test-failures for all tests but CLAMR where I couldn't figure out what the problem actually is and the code in the upstream repo seems substantially different now.

Most fixes should be fine as is but in `SingleSource/UnitTests/matrix-types-specs.cpp` I changed the type of a matrix to use signed integers instead of unsigned but it seems that these are only transpose tests so it should be fine?

Ping for potential review @nikic @fhahn

### DIFF
--- a/SingleSource/UnitTests/Vectorizer/common.h
+++ b/SingleSource/UnitTests/Vectorizer/common.h
@@ -66,10 +66,18 @@ static std::mt19937 rng;
 // Initialize arrays A with random numbers.
 template <typename Ty>
 static void init_data(const std::unique_ptr<Ty[]> &A, unsigned N) {
-  std::uniform_int_distribution<uint64_t> distrib(
-      std::numeric_limits<Ty>::min(), std::numeric_limits<Ty>::max());
-  for (unsigned i = 0; i < N; i++)
-    A[i] = distrib(rng);
+  if constexpr (std::is_floating_point_v<Ty>) {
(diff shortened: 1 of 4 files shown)


SingleSource/UnitTests/Vectorizer/common.h
```diff
@@ -66,10 +66,18 @@ static std::mt19937 rng;
...
-      std::numeric_limits<Ty>::min(), std::numeric_limits<Ty>::max());
-  for (unsigned i = 0; i < N; i++)
-    A[i] = distrib(rng);
+  if constexpr (std::is_floating_point_v<Ty>) {
+    std::uniform_real_distribution<Ty> distrib(
+        std::numeric_limits<Ty>::min(), std::numeric_limits<Ty>::max());
+    for (unsigned i = 0; i < N; i++)
+      A[i] = distrib(rng);
+  } else {
+    using DistTy = std::conditional_t<std::is_signed_v<Ty>, int64_t, uint64_t>;
```
nikic: Can we directly use `Ty` here? It seems pointless to generate in a larger type and then truncate...

T0b1-iOS: Yeah thank you. That was a leftover from a previous attempt at solving the failures.

PR #260, opened 2025-06-18
[utils] Migrate compare.py shebang to python3

tomershafir:
Many modern environments link `python` to Python 2 and `python3` to Python 3, so on those the script would be ran with Python 2. However, the script uses `pandas` and `scipy` which have long dropped support for Python 2. In addition, the code is effectively Python 3. Thus, its better to upgrade and explictly run the script using `python3`.


MatzeB:
LGTM

PR #261, opened 2025-06-19
[HashRecognize] Add CRC optz tests with type-mismatch

artagnon:


pfusik (changes requested):
These new tests seem to be copy-pasted with some minor changes.
Can we de-duplicate this code with templates or preprocessor?

artagnon:
> These new tests seem to be copy-pasted with some minor changes. Can we de-duplicate this code with templates or preprocessor?

Thanks, I'd been meaning to do this, and finally got around to it: https://github.com/llvm/llvm-test-suite/pull/262. To keep the tests as simple as possible, I've stuck to C pre-processor macros.

pfusik (approved):
LGTM

PR #264, opened 2025-06-25
[UnitTests] Add initial set of dedicated early-exit unit tests.

fhahn:
Adds initial unit tests for early-exit vectorization covering a variation of auto-vectorization and forced interleaving with pragmas.

The interleaving variant is currently mis-compiled and needs
 * https://github.com/llvm/llvm-project/pull/145340
 * https://github.com/llvm/llvm-project/pull/145394.

We should probably extend the tests to make sure we cover various other scenarios, including returning the loaded element for the early exit, different index types and array sizes.


SingleSource/UnitTests/Vectorizer/CMakeLists.txt
```diff
@@ -1,3 +1,12 @@
+
+# Enable matrix types extension tests for compilers supporting -fenable-matrix.
```
david-arm: I think the comment doesn't match the code below?

SingleSource/UnitTests/Vectorizer/early-exit.cpp
```diff
@@ -0,0 +1,100 @@
...
+    Init II(Src);                                                              \
+    _Pragma("clang loop vectorize(disable) interleave_count(1)") Loop          \
+  };                                                                           \
+  auto VectorFn = [](std::function<void(int *)> II) -> unsigned {              \
+    Init II(Src);                                                              \
+    _Pragma("clang loop vectorize(enable)") Loop                               \
+  };                                                                           \
+  auto InterleavedFn = [](std::function<void(int *)> II) -> unsigned {         \
+    Init II(Src);                                                              \
+    _Pragma("clang loop vectorize(enable) interleave_count(4)") Loop           \
```
david-arm, line 93: Looks like this PR cannot be merged until the pragma has handled correctly.

SingleSource/UnitTests/Vectorizer/early-exit.cpp
```diff
@@ -0,0 +1,100 @@
...
+  }
+}
+
+/// Define 3 test functions
+/// * one with vectorization and interleaving disabled,
+/// * one with auto-vectorization w/o any pargmas
+/// * one where VF is picked automatically and interleaving is forced.
+#define DEFINE_SCALAR_AND_VECTOR_EARLY_EXIT(Init, Src, Loop)                   \
+  auto ScalarFn = [](std::function<void(int *)> II) -> unsigned {              \
+    Init II(Src);                                                              \
```
david-arm: This is pretty confusing as it looks like constructing a variable. Can you split this out over different lines so it's clearer, i.e.

```
  InitSrc;
  II(Src);
```

SingleSource/UnitTests/Vectorizer/early-exit.cpp
```diff
@@ -0,0 +1,100 @@
...
+  std::array Tests = {std::make_pair(VectorFn, "autovec"),
+                      std::make_pair(InterleavedFn, "interleave-forced")};
+
+  // Check finding the target element at all indices between 0 and 512.
+  for (unsigned IdxToFind = 0; IdxToFind < 512; ++IdxToFind) {
+    // Labmda to initialize all array elements to one, except the one to look
+    // for to zero.
+    auto Init1 = [IdxToFind](int *Arr) {
+      std::fill_n(Arr, N, 1);
+      if (IdxToFind < N)
```
david-arm, line 36: Is there a risk that a future compiler optimisation basically figures out the answer based on this init code (after inlining) and just optimises `ScalarFn` to return a constant before we even reach the loop vectoriser? It might be better to have Init1 and Init2 as real functions marked as `noinline`.

SingleSource/UnitTests/Vectorizer/CMakeLists.txt
fhahn: Removed now, givne that it has been enabled by default for a while now

SingleSource/UnitTests/Vectorizer/early-exit.cpp
fhahn: In other places we prevent this fro caling through a function marked as `optnone`. Done the same thing here, thanks

SingleSource/UnitTests/Vectorizer/early-exit.cpp
fhahn: updated thanks

SingleSource/UnitTests/Vectorizer/early-exit.cpp
fhahn: This should also be working now as expected

fhahn:
ping

SingleSource/UnitTests/Vectorizer/early-exit.cpp
```diff
@@ -0,0 +1,109 @@
...
+    Init;                                                                      \
+    II(Src);                                                                   \
+    _Pragma("clang loop vectorize(disable) interleave_count(1)") Loop          \
+  };                                                                           \
+  auto VectorFn = [](std::function<void(int *)> II) -> unsigned {              \
+    Init;                                                                      \
+    II(Src);                                                                   \
+    _Pragma("clang loop vectorize(enable)") Loop                               \
+  };                                                                           \
+  auto InterleavedFn = [](std::function<void(int *)> II) -> unsigned {         \
```
david-arm, line 90: Is it worth having a variant of this where IC=4, VF=1 to test the interleaved/scalar case?

SingleSource/UnitTests/Vectorizer/early-exit.cpp
david-arm: Great! Is there a way to tell that these loops did actually vectorise? It's pretty hard to encourage loop vectorisation for early exit loops now with the dereferenceable SCEV checking code having to ensure that the pointers never wrap. I wouldn't be surprised if additional builtins are required to provide attributes to the pointers.

SingleSource/UnitTests/Vectorizer/early-exit.cpp
fhahn: Added, thanks

SingleSource/UnitTests/Vectorizer/early-exit.cpp
fhahn: Unfortuantely I don't think there's a way to test if they are actually vectorized expect checking the generated code or remarks. 

the current version uses a local array on the stack where the size is directly related to the bounds, which should be the most reliable way we have currently. We could also add ones using the dereferenceable builtins as follow-ups and possible ones without static info to make sure we also have coverage once first-faulting loads are supported

SingleSource/UnitTests/Vectorizer/early-exit.cpp
david-arm: Thanks for updating! Although I actually meant something like this

```
  _Pragma("clang loop vectorize_width(1) interleave_count(4)") Loop
```

. I think both this new one and the one below could still vectorise with VF>1 right?

fhahn: yep, added an interleaved-only variant as well, thanks

david-arm (approved):
LGTM!

fhahn:
@luporl thanks for the heads up, reverted for now, will take a closer look tomorrow morning

fhahn:
> Also failed in staging: https://lab.llvm.org/staging/#/builders/49/builds/1274

This failure looks puzzling and more like an issue with the build directory being stale unless I am missing something, as i t fails with 
```
FAIL: test-suite :: SingleSource/UnitTests/Vectorizer/early-exit.test (2324 of 2465)
******************** TEST 'test-suite :: SingleSource/UnitTests/Vectorizer/early-exit.test' FAILED ********************
/srv/bbworker/polly-x86_64-fdcserver/polly-x86_64-linux-test-suite/build/testsuite.build/tools/timeit-target --timeout 7200 --limit-core 0 --limit-cpu 7200 --limit-file-size 209715200 --limit-rss-size 838860800 --append-exitstatus --redirect-output /srv/bbworker/polly-x86_64-fdcserver/polly-x86_64-linux-test-suite/build/testsuite.build/SingleSource/UnitTests/Vectorizer/Output/early-exit.test.out --redirect-input /dev/null --chdir /srv/bbworker/polly-x86_64-fdcserver/polly-x86_64-linux-test-suite/build/testsuite.build/SingleSource/UnitTests/Vectorizer --summary /srv/bbworker/polly-x86_64-fdcserver/polly-x86_64-linux-test-suite/build/testsuite.build/SingleSource/UnitTests/Vectorizer/Output/early-exit.test.time /srv/bbworker/polly-x86_64-fdcserver/polly-x86_64-linux-test-suite/build/testsuite.build/SingleSource/UnitTests/Vectorizer/early-exit
/srv/bbworker/polly-x86_64-fdcserver/polly-x86_64-linux-test-suite/build/testsuite.build/tools/fpcmp-target /srv/bbworker/polly-x86_64-fdcserver/polly-x86_64-linux-test-suite/build/testsuite.build/SingleSource/UnitTests/Vectorizer/Output/early-exit.test.out /srv/bbworker/polly-x86_64-fdcserver/polly-x86_64-linux-test-suite/build/testsuite.build/SingleSource/UnitTests/Vectorizer/early-exit.reference_output
+ /srv/bbworker/polly-x86_64-fdcserver/polly-x86_64-linux-test-suite/build/testsuite.build/tools/fpcmp-target /srv/bbworker/polly-x86_64-fdcserver/polly-x86_64-linux-test-suite/build/testsuite.build/SingleSource/UnitTests/Vectorizer/Output/early-exit.test.out /srv/bbworker/polly-x86_64-fdcserver/polly-x86_64-linux-test-suite/build/testsuite.build/SingleSource/UnitTests/Vectorizer/early-exit.reference_output
/srv/bbworker/polly-x86_64-fdcserver/polly-x86_64-linux-test-suite/build/testsuite.build/tools/fpcmp-target: error: unable to open '/srv/bbworker/polly-x86_64-fdcserver/polly-x86_64-linux-test-suite/build/testsuite.build/SingleSource/UnitTests/Vectorizer/early-exit.reference_output'
```

fhahn:
> @fhahn You are right, the buildbot do not clean the test-suite's build dir, meaning that that *.test files, even if the test disappeared from test-suite, were kept in the build-dir. Fixed in [llvm/llvm-project@e509974](https://github.com/llvm/llvm-project/commit/e509974d423f1f260d63ec374463b5f75b18625e)

Great thanks!

I re-landed the PR, but with forced interleaving left out for now until https://github.com/llvm/llvm-project/issues/178644 is fixed so we can make progress with https://github.com/llvm/llvm-test-suite/pull/325

PR #266, opened 2025-07-17
[UnitTest] Add test for fmax reductions without fast-math.

fhahn:
Add unit tests for vectorizing FMax reductions without fast-math flags, covering a wide range of inputs, including various combinations of NaNs and signed-zeros.

### DIFF
--- a/SingleSource/UnitTests/Vectorizer/fmax-reduction.cpp
+++ b/SingleSource/UnitTests/Vectorizer/fmax-reduction.cpp
@@ -1,0 +1,218 @@
+#include <algorithm>
+#include <functional>
+#include <iostream>
+#include <limits>
+#include <memory>
+#include <stdint.h>
+
+#include "common.h"
+
+static bool isEqual(float A, float B) {
+  if (std::isnan(A))
+    return std::isnan(B);
+
+  if (A == 0.0f)
+    return std::signbit(A) == std::signbit(B);
+
+  return A == B;
+}
+
+template <typename Ty> using Fn1Ty = std::function<Ty(Ty *, unsigned)>;
+
+template <typename Ty>
+static void check(Fn1Ty<Ty> ScalarFn, Fn1Ty<Ty> VectorFn, float *Src,
+                  unsigned N, const char *Type) {
+  auto Reference = ScalarFn(Src, N);
+  auto ToCheck = VectorFn(Src, N);
+  if (!isEqual(Reference, ToCheck)) {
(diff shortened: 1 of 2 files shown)


SingleSource/UnitTests/Vectorizer/fmax-reduction.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+    // Check with multiple NaNs at different offsets.
+    for (unsigned Offset = 1; Offset != 16; ++Offset) {
+      if (Idx + Offset < N)
+        Src1[Idx + Offset] = std::numeric_limits<float>::quiet_NaN();
+
+      check(ScalarFn, VectorFn, &Src1[0], N, "full-with-multiple-nan");
+    }
+  }
+}
```
arsenm, line 154: It would be good to test signaling nans, but it will be busted all over the place 

Test some denormals?

SingleSource/UnitTests/Vectorizer/fmax-reduction.cpp
```diff
@@ -0,0 +1,218 @@
...
+#include <memory>
+#include <stdint.h>
+
+#include "common.h"
+
+static bool isEqual(float A, float B) {
+  if (std::isnan(A))
+    return std::isnan(B);
+
+  if (A == 0.0f)
```
arsenm: I think if `isnan(B)` and `A == 0.0f` this becomes dependent on the nan signbit

fhahn: I updated the check above to check if either is `nan`, in that case both must be nan to be equal. I also check here if B is zero. I think that should take care of that.

SingleSource/UnitTests/Vectorizer/fmax-reduction.cpp
fhahn: I added some tests with denormals both as start value and in the inputs, thanks

At least for AArch64, the tests also pass when replacing all quiet nans with signaling nans with https://github.com/llvm/llvm-project/pull/148239, as it just checks if it matches the behavior of the scalar loop.

But in general it may be a bit risky to check signaling NaNs, as the behavior may not be 100% consistent across platforms?

arsenm: It should be, but the state is in flux and has never been consistent

arsenm (approved):

PR #269, opened 2025-07-21
[AArch64] Skip test aarch64-init-cpu-features if FMV runtime unavailable.

IamYJLee:
## ERROR CONTEXT
Before this patch, LNT test runs on AArch64 platforms (e.g., macOS arm64) could fail with missing symbol errors:
```
--- Tested: 140 tests --
FAIL: SingleSource/Benchmarks/Misc/aarch64-init-cpu-features.compile_time (1 of 140)
FAIL: SingleSource/Benchmarks/Misc/aarch64-init-cpu-features.execution_time (29 of 140)
Processing Times
```
```
Undefined symbols for architecture arm64:
  "_RUNTIME_INIT", referenced from:
      _main in aarch64-init-cpu-features.llvm.o
ld: symbol(s) not found for architecture arm64
```

## Patch Description
- Selects __init_cpu_features_resolver for Darwin, and __init_cpu_features for Linux.
- Attempts to compile a dummy program to detect availability of the symbol via TARGET_LLVMGCC.
- Adds corresponding macro (HAS_DARWIN_FMV or HAS_LINUX_FMV) if detection succeeds.
- Skips test if the symbol is undefined.

This change fixes build errors when running LNT tests on AArch64 using Make rather than CMake.
This patch reflects the changes from [llvm/llvm-test-suite@d24d5dcd](https://github.com/llvm/llvm-test-suite/commit/d24d5dcdcbb748c52747823a0d3df20864fad600).

Co-authored-by: Lee Young Joon <dog3hk.dev@gmail.com>
Co-authored-by: Alexandros Lamprineas <alexandros.lamprineas@arm.com>
Co-authored-by: Momchil Velikov <momchil.velikov@arm.com>

### DIFF
--- a/SingleSource/Benchmarks/Misc/Makefile
+++ b/SingleSource/Benchmarks/Misc/Makefile
@@ -13,24 +13,12 @@ endif
 endif
 
 ifeq ($(ARCH),AArch64)
-	ifeq ($(TARGET_OS),Darwin)
-		CHECK_FMV := $(shell echo "void __init_cpu_features_resolver(void);" > CheckHasAArch64FMV.h && \
-			$(CC) -c -x c CheckHasAArch64FMV.h -o CheckHasAArch64FMV.o 2>/dev/null && \
-			echo yes || echo no; \
-			rm -f CheckHasAArch64FMV.h CheckHasAArch64FMV.o)
-		ifeq ($(CHECK_FMV),yes)
-			CFLAGS += -DHAS_DARWIN_FMV
-		endif
-	endif
-	ifeq ($(TARGET_OS),Linux)
-		CHECK_FMV := $(shell echo "void __init_cpu_features(void);" > CheckHasAArch64FMV.h && \
-			$(CC) -c -x c CheckHasAArch64FMV.h -o CheckHasAArch64FMV.o 2>/dev/null && \
-			echo yes || echo no; \
-			rm -f CheckHasAArch64FMV.h CheckHasAArch64FMV.o)
-		ifeq ($(CHECK_FMV),yes)
-			CFLAGS += -DHAS_LINUX_FMV
-		endif
-	endif
+ifeq ($(TARGET_OS),Darwin)
+CFLAGS += -DHAS_DARWIN_FMV
+endif
+ifeq ($(TARGET_OS),Linux)
+CFLAGS += -DHAS_LINUX_FMV	
+endif
 endif
 
 ifeq ($(ARCH),Mips)
--- a/SingleSource/Benchmarks/Misc/Makefile
+++ b/SingleSource/Benchmarks/Misc/Makefile
@@ -14,10 +14,24 @@ endif
 
 ifeq ($(ARCH),AArch64)
 ifeq ($(TARGET_OS),Darwin)
-CFLAGS += -DHAS_DARWIN_FMV
+SYMBOL = __init_cpu_features_resolver
 endif
 ifeq ($(TARGET_OS),Linux)
-CFLAGS += -DHAS_LINUX_FMV	
+SYMBOL = __init_cpu_features
+endif
+# TARGET_LLVMGCC refers to the compiler specified with the --cc option,
+# and can be checked in the build-tools.log file in the test results directory.
+SYMBOL := $(shell echo "extern void $(SYMBOL)(void); void (*p)() = $(SYMBOL); int main() {}" | \
+       $(TARGET_LLVMGCC) --rtlib=compiler-rt -x c - -o /dev/null 2>/dev/null && \
+       echo $(SYMBOL) || echo undefined)
+ifeq ($(SYMBOL),__init_cpu_features_resolver)
+CFLAGS += --rtlib=compiler-rt -DHAS_DARWIN_FMV
+endif
+ifeq ($(SYMBOL),__init_cpu_features)
+CFLAGS += --rtlib=compiler-rt -DHAS_LINUX_FMV
+endif
+ifeq ($(SYMBOL),undefined)
+PROGRAMS_TO_SKIP := aarch64-init-cpu-features
 endif
 endif
 
--- a/SingleSource/Benchmarks/Misc/Makefile
+++ b/SingleSource/Benchmarks/Misc/Makefile
@@ -12,6 +12,15 @@ PROGRAMS_TO_SKIP := dt
 endif
 endif
 
+ifeq ($(ARCH),AArch64)
+ifeq ($(TARGET_OS),Darwin)
+CFLAGS += -DHAS_DARWIN_FMV
+endif
+ifeq ($(TARGET_OS),Linux)
+CFLAGS += -DHAS_LINUX_FMV	
+endif
+endif
+
 ifeq ($(ARCH),Mips)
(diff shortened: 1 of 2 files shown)


IamYJLee:
@labrinea Thank you for your comment.

I've confirmed how `check_symbol_exists` functions and understand that `HAS_AARCH64_FMV` is not always true. In my environment, when I configure the project using `CMakeLists.txt`, `HAS_AARCH64_FMV` is set to False.

However, when I run LNT with a Makefile that includes my modifications, the build and execution complete without any issues.

In other words, even though `check_symbol_exists` confirms that the `__init_cpu_features_resolver` symbol is missing, there are no problems compiling and running `aarch64-init-cpu-features.c`.

IamYJLee:
@labrinea 
I'm using the following command and encountering an `Undefined Symbol` error. 

```
./mysandbox/bin/lnt runtest nt --sandbox ./result --cc ./llvm-project/build/bin/clang --test-suite ./llvm-test-suite -j 16 --only-test SingleSource/Benchmarks/Misc --isysroot `xcrun --show-sdk-path`
``` 

To match the `CMakeLists.txt` configuration and resolve this, I've made the following changes:

1. I added an `#if defined()` directive to `aarch64-init-cpu-features.c` to prevent the problematic code from being compiled when `HAS_***_FW` is False. (My initial attempt to use `PROGRAM_TO_SKIP` to exclude the file did not work.)

2. To replicate the logic from the `CMakeLists.txt` (below code), I added a step that generates and compiles a temporary header file. The result of this check is stored in the `CHECK_FMV` variable.
https://github.com/llvm/llvm-test-suite/blob/299a0b35ffae204ce8b1afa42c1b7646d54c948a/SingleSource/Benchmarks/Misc/CMakeLists.txt#L39-L40

## Testing

In my local environment, the check for the `__init_cpu_features*` symbol always results in 'no'. I've confirmed my changes work as intended by manually setting the `CHECK_FMV` variable to both 'yes' and 'no'.

Could you please test these changes in an environment where the `__init_cpu_features*` symbol check would result in 'yes'?

labrinea:
Hi @IamYJLee. Apologies for not guiding you how to improve this PR instead. Since I spent the time investing this issue from my end I thought it's easier to raise a new PR with my changes. Allow me to explain why this patch doesn't work.

* When overriding the compiler with `--cc` on the lnt command, it sets CC=None and Target_LLVMGCC=/path/to/compiler, so the shell compile command had no effect 
* Assuming the compiler was correctly set, the program you are trying to compile doesn't not attempt to find the definition of the symbol, so it always succeeds. Therefore I added an assignment to a function pointer. The compiler needs to be able to resolve this. If not you'll get error in which case the script returns something else than the symbol name.
* When not intending to run a test ifdef-ing out the source is a bad practice. Instead we want to use the Makefile variable PROGRAMS_TO_SKIP

Feel free to adjust this PR accordingly and credit @momchil-velikov for helping me on this on the commit message. There are still two issues with my patch that need to be addressed, so we need someone to review these first.

SingleSource/Benchmarks/Misc/Makefile
```diff
@@ -12,6 +12,29 @@ PROGRAMS_TO_SKIP := dt
...
+ifeq ($(TARGET_OS),Darwin)
+SYMBOL = __init_cpu_features_resolver
+endif
+ifeq ($(TARGET_OS),Linux)
+SYMBOL = __init_cpu_features
+endif
+# TARGET_LLVMGCC refers to the compiler specified with the --cc option,
+# and can be checked in the build-tools.log file in the test results directory.
+SYMBOL := $(shell echo "extern void $(SYMBOL)(void); void (*p)() = $(SYMBOL); int main() {}" | \
+       $(TARGET_LLVMGCC) --rtlib=compiler-rt -x c - -o /dev/null 2>/dev/null && \
```
IamYJLee, line 25: `TARGET_LLVMGCC` refers to the compiler specified with the --cc option, and can be checked in the `build-tools.log` file in the test results directory.

IamYJLee:
@jroelofs @labrinea @momchil-velikov Could you please take a look? Thanks!

SingleSource/Benchmarks/Misc/Makefile
IamYJLee, line 25: I reviewed the `build/CMakeFiles/CMakeConfigureLog.yaml` to trace how `check_symbol_exists` is executed through `CMakeLists.txt`.
`compiler-rt` (a runtime library) is not shown in the configuration log, but including it in the current patch appears to work fine without introducing any issues.
Here is `check_symbol_exists` Log :
```
buildResult:
      variable: "HAS_AARCH64_FMV"
      cached: true
      stdout: |
        Change Dir: /Users/nshc-yjlee_m/Documents/dev/llvm-test-suite/build/CMakeFiles/CMakeScratch/TryCompile-j5irsv
        
        Run Build Command(s):/opt/homebrew/bin/ninja -v cmTC_f539e && [1/2] /Applications/Xcode161_Testing.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc   -arch arm64 -isysroot /Applications/Xcode161_Testing.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.1.sdk -MD -MT CMakeFiles/cmTC_f539e.dir/CheckSymbolExists.c.o -MF CMakeFiles/cmTC_f539e.dir/CheckSymbolExists.c.o.d -o CMakeFiles/cmTC_f539e.dir/CheckSymbolExists.c.o -c /Users/nshc-yjlee_m/Documents/dev/llvm-test-suite/build/CMakeFiles/CMakeScratch/TryCompile-j5irsv/CheckSymbolExists.c
        [2/2] : && /Applications/Xcode161_Testing.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -arch arm64 -isysroot /Applications/Xcode161_Testing.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.1.sdk -Wl,-search_paths_first -Wl,-headerpad_max_install_names  CMakeFiles/cmTC_f539e.dir/CheckSymbolExists.c.o -o cmTC_f539e   && :
        
      exitCode: 0
```

IamYJLee:
> As I said earlier I am not sure if TARGET_LLVMGCC is the right compiler and whether --rtlib=compiler-rt should be inherited from some CFLAGS variable. I tested this patch and it worked in my environment.

The logs for `check_symbol_exists` indicate that the symbol detection is being carried out using the system’s default compiler rather than the clang compiler we aim to test.
This raises the possibility that `CC` might be the more appropriate variable to use instead of `TARGET_LLVMGCC`.

labrinea:
I am afraid I will only be able to look into this further, after I am back from vacation in two weeks time.

IamYJLee:
@labrinea 
PING: Would it be okay to merge this PR first and handle any additional changes in a follow-up?
Please let me know if there’s anything that needs to be addressed before merging.

IamYJLee:
@labrinea I’d appreciate it if you could merge this 🙂.

labrinea:
Do you not have commit access?

PR #270, opened 2025-07-24
[FMV][AArch64] Skip test aarch64-init-cpu-features if runtime unavailable

labrinea:
When we use make to build the llvm-test-suite instead of cmake
the test aarch64-init-cpu-features should be skipped unless one
of __init_cpu_features_resolver or __init_cpu_features is found.

Co-authored-by: Momchil Velikov <momchil.velikov@arm.com>

### DIFF
--- a/SingleSource/Benchmarks/Misc/Makefile
+++ b/SingleSource/Benchmarks/Misc/Makefile
@@ -23,4 +23,25 @@ RUNTIMELIMIT:=10800
 endif
 endif
 
+ifeq ($(ARCH),AArch64)
+ifeq ($(TARGET_OS),Darwin)
+SYMBOL = __init_cpu_features_resolver
+endif
+ifeq ($(TARGET_OS),Linux)
+SYMBOL = __init_cpu_features
+endif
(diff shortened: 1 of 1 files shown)


SingleSource/Benchmarks/Misc/Makefile
```diff
@@ -23,4 +23,25 @@ RUNTIMELIMIT:=10800
...
 
+ifeq ($(ARCH),AArch64)
+ifeq ($(TARGET_OS),Darwin)
+SYMBOL = __init_cpu_features_resolver
+endif
+ifeq ($(TARGET_OS),Linux)
+SYMBOL = __init_cpu_features
+endif
+SYMBOL := $(shell echo "extern void $(SYMBOL)(void); void (*p)() = $(SYMBOL); int main() {}" | \
+       $(TARGET_LLVMGCC) --rtlib=compiler-rt -x c - -o /dev/null 2>/dev/null && \
```
labrinea, line 34: I am not sure whether `$(TARGET_LLVMGCC)` is the right compiler. Also whether `--rtlib=compiler-rt` should be inferred from the existing CFLAGS

PR #273, opened 2025-07-31
Rename reference_output file (NFC)

kkwli:
`llvm_test_traditional()` looks for the reference output filename without `.cpp`.

https://github.com/llvm/llvm-test-suite/blob/eed578a4536b25b7e30ce047b887698e65d5e9f3/cmake/modules/SingleMultiSource.cmake#L112-L124


kkwli:
Ping. Any comments are welcome. Thanks.

PR #275, opened 2025-08-13
Support collecting perf profiles remotely

lukel97:
If we're running tests on a remote machine with -DTEST_SUITE_REMOTE_HOST and generate a perf profile with `lit --param profile=perf`, currently the profile step will be run on the local machine and fail.

This patch fixes this by running the profile step on the remote, as well as copying the generated profiles back to the host so tools like LNT can inspect them.

There's two parts to this, first we need to run the remote module after the perf (and hpmcount) module so that we can wrap the profile step with ssh etc.

Secondly, we need to scp over the generated .perf_data files like we also do for the profilegen.py module. However today this is also done as part of the profile step which we now need to start running over ssh. So to fix this, this adds a new "profile collect" step that runs after the profile step, but always runs on the local machine.

I've tested this out on a remote build configured with `-DTEST_SUITE_REMOTE_HOST=... -DTEST_SUITE_USE_PERF=ON -DTEST_SUITE_PROFILE_GENERATE=ON -DTEST_SUITE_USE_IR_PGO=ON` and running the tests with `lit --param profile=perf`, and it seems to work.

### DIFF
--- a/lit.cfg
+++ b/lit.cfg
@@ -25,11 +25,6 @@ config.traditional_output = False
 config.single_source = False
 if "SSH_AUTH_SOCK" in os.environ:
     config.environment["SSH_AUTH_SOCK"] = os.environ["SSH_AUTH_SOCK"]
-if not hasattr(config, "remote_host"):
-    config.remote_host = ""
-config.remote_host = lit_config.params.get("remote_host", config.remote_host)
-if config.remote_host:
-    config.test_modules.append("remote")
 
 # Load previous test results so we can skip tests that did not change.
 previous_results_file = lit_config.params.get("previous", None)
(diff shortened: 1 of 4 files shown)


lukel97:
Gentle ping, I'm aware that this probably isn't an area that people look at often so apologies in advance!

asb (approved):
My familiarity with the code is low, but your description makes sense and I've stepped through the changes in context, and they seem sensible to me.

jroelofs (approved):
LGTM. Sorry for the delay.

lukel97:
> LGTM. Sorry for the delay.

Not at all, thanks for taking the time to take a look :)

PR #276, opened 2025-08-14
[HIP] Add a __builtin_memset test

ritter-x2a:
This is mostly analogous to the existing memmove test: __builtin_memset is called on the device with different lengths, in different address spaces, at different dynamic alignments, and with constant and dynamic operands. The result is compared to a reference computation on the host to ensure functional correctness of the memset lowering.

Also adds a missing standard lib include to the memmove test.

For SWDEV-543208.

### DIFF
--- a/External/HIP/memset.hip
+++ b/External/HIP/memset.hip
@@ -1,0 +1,377 @@
+#include <cassert>
+#include <cstdint>
+#include <cstring>
+#include <iostream>
+#include <memory>
+#include <vector>
+
+#include "hip/hip_runtime.h"
+
+// Tests for the functional correctness of the lowering of memset in device
+// code. Various memsets are performed on device side and the result of each is
+// compared to the corresponding operation on the host. Global, shared, and
+// stack memory is tested.
+
+#define VERBOSE 0
+
+#define CHKHIP(r)                                                              \
+  if (r != hipSuccess) {                                                       \
+    std::cerr << hipGetErrorString(r) << std::endl;                            \
+    abort();                                                                   \
+  }
+
+// Maximal number of bytes to set with a memset call, used to allocate
+// buffers.
+#define MAX_BYTES_PER_THREAD 2048
+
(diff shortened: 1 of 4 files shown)


jplehr (approved):
lgtm

External/HIP/memset.hip
```diff
@@ -0,0 +1,377 @@
...
+
+#define CHKHIP(r)                                                              \
+  if (r != hipSuccess) {                                                       \
+    std::cerr << hipGetErrorString(r) << std::endl;                            \
+    abort();                                                                   \
+  }
+
+// Maximal number of bytes to set with a memset call, used to allocate
+// buffers.
+#define MAX_BYTES_PER_THREAD 2048
```
arsenm: Maybe static constexpr variables instead of old fashioned macros?

External/HIP/memset.hip
```diff
@@ -0,0 +1,377 @@
...
+                                    : (memset_kernel_global<SZ, SetVal, true,
+                                                            false, false>)))
+             : (use_tid
+                    ? (memset_kernel_global<SZ, SetVal, false, true, false>)
+                    : (const_setval ? (memset_kernel_global<SZ, SetVal, false,
+                                                            false, true>)
+                                    : (memset_kernel_global<SZ, SetVal, false,
+                                                            false, false>)))),
+        dim3(num_blocks), dim3(block_size), 0, 0, buf_device, dst_idx, SetVal,
+        SZ);
```
arsenm: This is too many levels of nested ternary operators, can you break this up some?

External/HIP/memset.hip
ritter-x2a: Done.

External/HIP/memset.hip
ritter-x2a: I think I've simplified the kernel selection now, but feel free to suggest more improvements.

PR #277, opened 2025-08-14
[flang][tests] Disable Fortran/gfortran/regression/finalize_38.f90

klausler:
One test case in finalize_38.f90 expects that an allocatable component of an expression result will not be finalized.  The flang-new compiler is being updated by the pull request

  https://github.com/llvm/llvm-project/pull/153509

to ensure finalization of allocatables when they are deallocated, since this is an event that triggers finalization (F'2023 7.5.6.3p2) on a finalizable (3.71) data entity (3.41).


Fortran/gfortran/regression/DisabledFiles.cmake
```diff
@@ -383,6 +383,9 @@ file(GLOB UNIMPLEMENTED_FILES CONFIGURE_DEPENDS
   norm2_3.f90
   pr96711.f90
 
+  # pmk?
```
eugeneepshteyn: This test was not in the description?

Fortran/gfortran/regression/DisabledFiles.cmake
```diff
@@ -1850,4 +1850,7 @@ file(GLOB FAILING_FILES CONFIGURE_DEPENDS
   # warning, so long as it is not used in that scope.
   use_15.f90
   use_rename_8.f90
+
+  # The "stop 171" case expects an incorrect number of finalizations.
+  finalize_38.f90
```
tarunprabhu: From the description of the PR, it sounds like this test will never pass with flang because, gfortran's finalization behavior is different from flang's. If this is the case, it may be better to move this to the `UNSUPPORTED_FILES` list above.

tarunprabhu (approved):
Thanks for the change.

PR #281, opened 2025-09-04
[flang][tests] add test for llvm-project#156753

akuhlens:
This test checks that child list IO can read past the end of a record as implemented by: https://github.com/llvm/llvm-project/pull/156753.


Fortran/UnitTests/defined_io/list-read.f90
```diff
@@ -0,0 +1,65 @@
+module m
```
klausler: Is this a verbatim copy of a test case in a bug report?  If so, please link to it in a source comment.

Fortran/UnitTests/defined_io/list-read.f90
```diff
@@ -0,0 +1,65 @@
...
+end module
+
+program list_read
+    use m
+    integer currentPos, ss
+
+    type(base) :: b1(2)
+
+
+    open (1, file='dcmlChildRead003.data', access='stream', form='formatted', &
```
klausler: This OPEN statement should use a file name that is test name dependent, and have a STATUS= specifier of 'scratch'.

akuhlens:
@tarunprabhu I just noticed that the reference output doesn't seem to be checked. How do I modify this PR to check the output of the program run?

tarunprabhu:
I just built this and the test fails. It looks like the reference output is being checked, but perhaps there is something else going on.

```
$ llvm-lit -v Fortran/UnitTests/defined_io/llvm-project-issue-154979.test

/llvm-test-suite/tools/fpcmp-target: Compared: 1.000000e+01 and 1.200000e+01
abs. diff = 2.000000e+00 rel.diff = 1.666667e-01
Out of tolerance: rel/abs: 1.000000e-03/0.000000e+00
```

akuhlens:
> I just built this and the test fails. It looks like the reference output is being checked, but perhaps there is something else going on.

I just realized there was an option in my build script that was preventing the tests from running: `-DTEST_SUITE_RUN_BENCHMARKS=OFF`

This test won't pass until the PR in the description is landed, but I can confirm it passes now.

akuhlens:
It is kinda wild that we don't run the Fortran test suite as a land time CI job when modifying it.

PR #282, opened 2025-09-05
[HIP] Removed STDLIB_VERSION Macro definition

Sandhyarani-Y:
Removed STDLIB_VERSION Macro definition to array.hip file

### DIFF
--- a/External/HIP/array.hip
+++ b/External/HIP/array.hip
@@ -14,18 +14,7 @@
 #include <array>
 #include <assert.h>
 
-// Check the C++ standard and define the library version accordingly
-#if defined(__cpp_lib_concepts) && (__cpp_lib_concepts >= 201811L)
-    #define STDLIB_VERSION 2020
-#elif defined(__cpp_lib_parallel_algorithm) && (__cpp_lib_parallel_algorithm >= 201603L)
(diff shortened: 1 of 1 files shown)


External/HIP/array.hip
```diff
@@ -14,6 +14,17 @@
...
+#if defined(__cpp_lib_concepts) && (__cpp_lib_concepts >= 201811L)
+    #define STDLIB_VERSION 2020
+#elif defined(__cpp_lib_parallel_algorithm) && (__cpp_lib_parallel_algorithm >= 201603L)
+    #define STDLIB_VERSION 2017
+#elif __cplusplus >= 201402L
+    #define STDLIB_VERSION 2014
+#else
+    #define STDLIB_VERSION 2011
+#endif
+
```
jmmartinez: Lets just assume STDLIB_VERSION matches the requested C++ standard; so lets remove the STDLIB_VERSION checks from this file.

Although good, this trick depends on including the appropriate c++ header (e.g. `__cpp_lib_parallel_algorithm` is not defined unless `<algorithm>` is included, even if it is supported); and when it doesn't work it fails silently.

jmmartinez (approved):

PR #285, opened 2025-09-26
Fix SPEC CPU 2017 benchmarks with GCC 15

lukel97:
4 of the SPEC CPU 2017 benchmarks fail to build with the latest GCC, see: https://cc-perf.igalia.com/db_default/v4/nts/31

526.blender_r, 544.nab_r and 502.gcc_r fail because GCC 15 changed the default C standard to gnu23, so this fixes it by telling CMake to use -std=gnu17, which matches Clang's default. I checked and even though CMake doesn't mention gnu17 specifically, it passes -std=gnu17. This is the C version of 7ebe5edc24

523.xalancbmk_r fails for a latent error in a template. We've already had to workaround it in Clang in 92b58b3bf4c6001bff4c7c7f8f98305e3ead56ff but GCC has it's own specific -Wno-error flag to ignore it, see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=116064

### DIFF
--- a/External/SPEC/CFP2017rate/526.blender_r/CMakeLists.txt
+++ b/External/SPEC/CFP2017rate/526.blender_r/CMakeLists.txt
@@ -1219,5 +1219,6 @@ speccpu2017_add_executable(
   specrand/specrand.c
   spec_backtrace.c
 )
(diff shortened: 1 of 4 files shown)


fhahn (approved):
LGTM, thanks

PR #287, opened 2025-10-01
[flang][runtime] add new test for extends_type_of

akuhlens:
Test program for https://github.com/llvm/llvm-project/issues/155459.


Fortran/UnitTests/descriptor/extends_type_of.f90
```diff
@@ -0,0 +1,100 @@
...
+  use extends_type_of_mod
+  type(p1), target :: p, r
+  type(p2), target :: q
+  type(k1(10)), target :: k10
+  type(k1(20)), target :: k20
+  integer, target :: i1, i2
+  integer(kind=8), target :: i3
+  real, target :: f
+  
+  class(*), pointer :: null_cp => null()
```
klausler, line 67: If you do similar testing with allocatables, I think you'll find the bug in the implementation.

akuhlens: Actually, mirroring the tests to allocatables didn't seem to turn anything up... Which case are you expecting to fail?

akuhlens: Maybe I recorded the wrong answer for a test.

klausler: Your original implementation used `mold` instead of `a` in a Boolean expression.

klausler (approved):

PR #345, opened 2026-02-23
[MicroBenchmarks] Add benchmark for control-flow-vectorization.

ElvisWang123:
Included benchmarks with conditional loops to trigger control-flow vectorization. 
These cases can be used for measuring the  performance impact of control-flow vectorization across targets.

### DIFF
--- a/MicroBenchmarks/LoopVectorization/ControlFlowVectorization.cpp
+++ b/MicroBenchmarks/LoopVectorization/ControlFlowVectorization.cpp
@@ -1,0 +1,188 @@
+#include <iostream>
+#include <memory>
+#include <random>
+
+#include "benchmark/benchmark.h"
+
+#define ITERATIONS 100000
+
+template <typename T> using CFVFunc = void (*)(T *, unsigned);
+
+// Define conditional increment loop with given stride.
+#define DEF_COND_INC_LOOP(name, stride)                                        \
+  template <typename T>                                                        \
+  __attribute__((noinline)) static void run_##name##_autovec(T *A,             \
+                                                             unsigned N) {     \
+    for (unsigned i = 0; i < N; i++) {                                         \
+      if (i % stride == 0) {                                                   \
+        A[i] = A[i] + 1;                                                       \
+      }                                                                        \
+    }                                                                          \
+  }                                                                            \
+  template <typename T>                                                        \
+  __attribute__((noinline)) static void run_##name##_novec(T *A, unsigned N) { \
+    _Pragma("clang loop vectorize(disable) interleave(disable)")               \
+    for (unsigned i = 0; i < N; i++) {                                         \
+      if (i % stride == 0) {                                                   \
+        A[i] = A[i] + 1;                                                       \
+      }                                                                        \
+    }                                                                          \
+  }
+
+// Define conditional increment by value loop.
+#define DEF_COND_INC_VALUE_LOOP(name, marker)                                  \
+  template <typename T>                                                        \
+  __attribute__((noinline)) static void run_##name##_autovec(T *A,             \
+                                                             unsigned N) {     \
+    for (unsigned i = 0; i < N; i++) {                                         \
+      if (A[i] == marker) {                                                    \
+        A[i] = A[i] + 1;                                                       \
+      }                                                                        \
+    }                                                                          \
+  }                                                                            \
+  template <typename T>                                                        \
+  __attribute__((noinline)) static void run_##name##_novec(T *A, unsigned N) { \
+    _Pragma("clang loop vectorize(disable) interleave(disable)")               \
+    for (unsigned i = 0; i < N; i++) {                                         \
+      if (A[i] == marker) {                                                    \
+        A[i] = A[i] + 1;                                                       \
+      }                                                                        \
+    }                                                                          \
+  }
+
+// Define unconditional increment loop.
+template <typename T>
+__attribute__((noinline)) static void run_uncond_inc_autovec(T *A, unsigned N) {
+  for (unsigned i = 0; i < N; i++) {
+    A[i] = A[i] + 1;
+  }
+}
+
+template <typename T>
+__attribute__((noinline)) static void run_uncond_inc_novec(T *A, unsigned N) {
+  _Pragma("clang loop vectorize(disable) interleave(disable)")
+  for (unsigned i = 0; i < N; i++) {
+    A[i] = A[i] + 1;
+  }
+}
+
+// Define loops with different strides.
+// stride=2: 50% active lanes
+// stride=4: 25% active lanes
+// stride=8: 12.5% active lanes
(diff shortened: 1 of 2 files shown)


MicroBenchmarks/LoopVectorization/ControlFlowVectorization.cpp
```diff
@@ -0,0 +1,188 @@
...
+DEF_COND_INC_LOOP(cond_inc_stride_2, 2)
+DEF_COND_INC_LOOP(cond_inc_stride_4, 4)
+DEF_COND_INC_LOOP(cond_inc_stride_8, 8)
+DEF_COND_INC_LOOP(cond_inc_stride_16, 16)
+DEF_COND_INC_LOOP(cond_inc_stride_32, 32)
+DEF_COND_INC_LOOP(cond_inc_stride_64, 64)
+DEF_COND_INC_LOOP(cond_inc_stride_128, 128)
+
+// Conditional increment by value (sparse condition).
+DEF_COND_INC_VALUE_LOOP(cond_inc_by_value, 42)
```
npanchen: what's going to be a % of active lanes here ?  Is it really worth to be tracked ?

MicroBenchmarks/LoopVectorization/ControlFlowVectorization.cpp
```diff
@@ -0,0 +1,188 @@
...
+#include "benchmark/benchmark.h"
+
+#define ITERATIONS 100000
+
+template <typename T> using CFVFunc = void (*)(T *, unsigned);
+
+// Define conditional increment loop with given stride.
+#define DEF_COND_INC_LOOP(name, stride)                                        \
+  template <typename T>                                                        \
+  __attribute__((noinline)) static void run_##name##_autovec(T *A,             \
```
npanchen: What is the meaning of this benchmark ? Just track current state of cf vectorization of novec and autovec  or help to identify better LMUL to vectorize the loop ? If latter, it does make sense to add similar functions with forced vectorization for default LMUL and specified LMULs

MicroBenchmarks/LoopVectorization/ControlFlowVectorization.cpp
ElvisWang123: Removed small-stride cases to focus on larger strides. 
This allows for entirely inactive lanes (in most of the cases), which helps testing conditional vector block optimizations(control-flow vectorization) across different targets.

MicroBenchmarks/LoopVectorization/ControlFlowVectorization.cpp
ElvisWang123: IIUC, this benchmark serves as a test suite for other targets to measure the performance impact of enabling control-flow vectorization.

I've updated the PR description to make it more clear.

MicroBenchmarks/LoopVectorization/ControlFlowVectorization.cpp
```diff
@@ -0,0 +1,177 @@
+#include <iostream>
+#include <memory>
+#include <random>
+
+#include "benchmark/benchmark.h"
+
+#define ITERATIONS 100000
+
+template <typename T> using CFVFunc = void (*)(T *, unsigned);
```
fhahn: might be good to use a more descriptive name or add a comment

MicroBenchmarks/LoopVectorization/ControlFlowVectorization.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+// Define conditional increment loop with given stride.
+#define DEF_COND_INC_LOOP(name, stride)                                        \
+  template <typename T>                                                        \
+  __attribute__((noinline)) static void run_##name##_autovec(T *A,             \
+                                                             unsigned N) {     \
+    for (unsigned i = 0; i < N; i++) {                                         \
+      if (i % stride == 0) {                                                   \
```
fhahn: Stride may be a bit mis-leading here, maybe to be confused with the stride by which pointers increment? `stride` here effectively controls how frequently the condition executes, right?

MicroBenchmarks/LoopVectorization/ControlFlowVectorization.cpp
```diff
@@ -0,0 +1,177 @@
...
+  _Pragma("clang loop vectorize(disable) interleave(disable)")
+  for (unsigned i = 0; i < N; i++) {
+    A[i] = A[i] + 1;
+  }
+}
+
+// Define loops with different strides.
+// Stride 16 usually big enough to accross single vector which can test if
+// control-flow-vectorization is profitable on these loops.
+DEF_COND_INC_LOOP(cond_inc_stride_16, 16)
```
fhahn: This variant roughly executes the conditional code every 16 iterations, right?

Would be good to also add variations where conditional code executes more frequently, including extreme case (every iteration, every other iteration)?

MicroBenchmarks/LoopVectorization/ControlFlowVectorization.cpp
ElvisWang123: Update to `ControlFlowLoopFunc`. Thanks!

MicroBenchmarks/LoopVectorization/ControlFlowVectorization.cpp
ElvisWang123: Yes. Update the naming to the `branch_every_N`. Thanks!

MicroBenchmarks/LoopVectorization/ControlFlowVectorization.cpp
ElvisWang123: Yes.

Add the cases that branch will taken for every (1, 2, 4, 8) iters. Thanks!

ElvisWang123:
Gentle ping :)

ElvisWang123:
Gentle ping:) @fhahn

MicroBenchmarks/LoopVectorization/ControlFlowVectorization.cpp
```diff
@@ -0,0 +1,181 @@
...
+        std::cerr << "ERROR: vectorization result different at index " << i
+                  << "; " << A_vec[i] << " != " << A_novec[i] << "\n";
+        exit(1);
+      }
+    }
+  }
+#endif
+
+  for (auto _ : state) {
+    std::copy(&A[0], &A[0] + ITERATIONS, &A_vec[0]);
```
fhahn: hmmm, would it be possible to move this out of the measurement loop, as this memcopy may introduce additional noise and means we are emeasuring more than the interesting loop.

Perhaps also worth adding a variant where the condition reads from a different area than the stores ?

ElvisWang123: Updated, thanks!

PR #351, opened 2026-02-25
Add gitignore entry for macOS DS Store files

tmjbios:
Just a simple addition of the .DS_Store entry to the gitignore.


carlocab (commented):
Not necessarily opposed to this, but this feels like something that belongs in your global gitignore instead.

jroelofs (approved):

PR #373, opened 2026-03-27
Add compile-time max RSS reporting support

tomershafir:
This patch adds basic compile/link time max RSS reporting to llvm-test-suite. Memory usage is an important metric for users, and thus also for LLVM developer to better understand the compile/link time impact of a compiler change.

Memory usage and its measurement in general can get very complex and platform specific. A simple but still useful aggregate is max working set size. It is reported by `getrusage()` on Apple platforms and Linux, which are the scope of this patch. Even though `getrusage()` is a POSIX API, the `maxrss` part of it is not and is platform dependent. Support for other platforms can be added later.

There is already an orthogonal reporting capabily by https://llvm-compile-time-tracker.com/ but adding this capability to llvm-test-suite can offer more programmability and integration opportunities.

Bottom-up implementation:
- Add a new `--report-maxrss` option to `timeit` which collects `maxrss` for the supported platforms from `getrusage()` and writes it to the summary file or the non-POSIX terminal
- Add LIT bindings for `maxrss` read from output to `litsupport/modules/timeit.py`
- Add max-based aggregation for `maxrss` to `litsupport/modules/compiletime.py`
- Add a CMake option `TEST_SUITE_REPORT_COMPILE_MAXRSS` that is only active under `TEST_SUITE_COLLECT_COMPILE_TIME`


tomershafir:
@nikic Hi, I wonder if you can take a look, given you experience with compile-time tracking

fhahn (commented):
Thanks for the patch! I think it would be good to add a test this is actually working as expected, similar to `tools/test/test_not.py`

I think it should be possible to  add a simple binary that just allocates X MB + calling memset on it, where X is provided as argument, then run a test to check the reported MAX_RSS matches roughly X.

tools/test/test_maxrss.py
```diff
@@ -0,0 +1,45 @@
+"""Test that timeit --report-maxrss reports a valid max RSS value.
+
+Usage: test_maxrss.py <timeit> <alloc_mem>
+"""
+import os
+import subprocess
+import sys
+import tempfile
+
+ALLOC_MiB = 32
```
fhahn: might be good to choose a higher value here to make more robust

fhahn (approved):
LGTM, thanks

luporl:
I have noticed that `tools/test/alloc_mem.test` sometimes fails on buildbots:
- https://lab.llvm.org/buildbot/#/builders/79/builds/3255
- https://lab.llvm.org/buildbot/#/builders/121/builds/2338
- https://lab.llvm.org/buildbot/#/builders/125/builds/14529

A common error message is
```
maxrss too low: 132120576 < 134217728
```

I don't know why this happens, but perhaps we could tolerate a lower reported value, such as 10% less than the requested allocation?

lukel97:
> I have noticed that `tools/test/alloc_mem.test` sometimes fails on buildbots:
> 
> * https://lab.llvm.org/buildbot/#/builders/79/builds/3255
> * https://lab.llvm.org/buildbot/#/builders/121/builds/2338
> * https://lab.llvm.org/buildbot/#/builders/125/builds/14529
> 
> A common error message is
> 
> ```
> maxrss too low: 132120576 < 134217728
> ```
> 
> I don't know why this happens, but perhaps we could tolerate a lower reported value, such as 10% less than the requested allocation?

I think getrusage is sampling based so it's not guaranteed to be accurate. I think for that test it's probably good enough to check that the maxrss > 0

tomershafir:
~~Interestingly, the dynamic max RSS being reported in all of the build failures above (left hand side of the comparison) is exactly the same. So this seems to be possibly a deterministic behavior we may be able to account for. However, I am not able to reproduce the issue on macOS machines.~~

~~I think the values are fairly close so we can take a first step to lower the threshold by a small percentage, rather than a much weaker `maxrss > 0` test that should mostly/always be satisfied.~~

~~Ill prepare a patch for tolerating a 10% diff.~~

~~@luporl @lukel97 If you are interested, feel free to investigate why it sometimes fail on Linux, that would be great :)~~
Rethinking, the max RSS number reported by the OS is generally not strictly reliable and can have exhibit non-deterministic behavior. Thus, we may only check conservatively that the number is just greater than 0.

PR #377, opened 2026-04-05
[compare.py] Add confidence interval

mustartt:
This patch adds a `--diff-confidence-interval=relative|absolute` option to `compare.py` to report 95% (or 1-alpha) confidence intervals for the relative difference between `lhs` and `rhs` runs.

The current p-value and significance markers only tells the user if a difference is statistically significant against null hypothesis, but does not show how large the true difference might vary.

Example output from `compare.py ... --statistics --diff-confidence-interval`
```
     Program                                       exec_time
                                                   lhs       rhs    diff  std_lhs std_rhs t-value p-value significant diff_ci_rel
     C                                               2.95      3.40 15.3% 0.076   0.100   -6.653  0.0027  Y           [ 9.3%, 22.7%]
     A                                               1.00      1.15 15.0% 0.050   0.050   -3.674  0.0213  Y           [ 3.5%, 25.1%]
     B                                               1.95      2.20 12.8% 0.076   0.050   -4.427  0.0114  Y           [ 4.3%, 18.8%]
                                Geomean difference                  14.4%
```

### DIFF
--- a/utils/compare.py
+++ b/utils/compare.py
@@ -128,7 +128,7 @@ def merge_values(values, merge_function):
 
 
 def get_values(values, lhs_name=None, rhs_name=None):
-    exclude_cols = ["diff", "t-value", "p-value", "significant"]
+    exclude_cols = ["diff", "t-value", "p-value", "significant", "ci_lower", "ci_upper"]
     exclude_cols.extend([f'std_{lhs_name}', f'std_{rhs_name}'])
     exclude_cols.extend([f'cv_{lhs_name}', f'cv_{rhs_name}'])
     values = values[[c for c in values.columns if c not in exclude_cols]]
@@ -181,9 +181,9 @@ def compute_statistics(lhs_d, rhs_d, metrics, alpha, coeff_var, lhs_name, rhs_na
             if len(lhs_values) >= 2 and len(rhs_values) >= 2:
                 lhs_std = lhs_values.std(ddof=1)
                 rhs_std = rhs_values.std(ddof=1)
+                lhs_mean = lhs_values.mean()
+                rhs_mean = rhs_values.mean()
                 if coeff_var:
-                    lhs_mean = lhs_values.mean()
-                    rhs_mean = rhs_values.mean()
                     stats_dict[metric][program] = {
                         f'cv_{lhs_name}': lhs_std / lhs_mean if lhs_mean != 0 else float('nan'),
                         f'cv_{rhs_name}': rhs_std / rhs_mean if rhs_mean != 0 else float('nan'),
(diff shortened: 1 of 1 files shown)


mustartt:
I believe this should be enabled by default? If there are any objections, it can be gated  under a flag too.

tomershafir:
I think we can compact the output a bit to a single column called `diff_ci` formatted as `[-0.1%, 0.1%]` for example, and save some space on the screen.

Similar concerns make me think we should gate the new column under a flag like `--confidence-interval` (not just a vauge `--ci`). E.g. llvm-test-suite reports 3 metrics by default (compile time, code size, runtime) which would report 3 extra relatively long columns to the screen (after compaction), possibly breaking the view.

utils/compare.py
```diff
@@ -128,7 +128,7 @@ def merge_values(values, merge_function):
 
 
 def get_values(values, lhs_name=None, rhs_name=None):
-    exclude_cols = ["diff", "t-value", "p-value", "significant"]
+    exclude_cols = ["diff", "t-value", "p-value", "significant", "diff_ci"]
```
tomershafir: If we already here, lets remove calls to `extend()` and just initialize an array with all the columns, `diff_ci` being last

utils/compare.py
```diff
@@ -163,7 +163,7 @@ def add_diff_column(metric, values, absolute_diff=False):
     return values
 
 
-def compute_statistics(lhs_d, rhs_d, metrics, alpha, coeff_var, lhs_name, rhs_name):
+def compute_statistics(lhs_d, rhs_d, metrics, alpha, coeff_var, lhs_name, rhs_name, confidence_interval=False):
```
tomershafir: - No need in default value, its already handled by the caller. 
- If we are already here, lets move `lhs_name` and `rhs_name` to the start of the param list after `*_d` so we can more structurally append new params
- Rethinking, I think we should explicitly denote its a diff CI, and not CI of something else, thus maybe `diff_conf_int` would be a better name?

utils/compare.py
```diff
@@ -193,10 +193,21 @@ def compute_statistics(lhs_d, rhs_d, metrics, alpha, coeff_var, lhs_name, rhs_na
                         f'std_{lhs_name}': lhs_std,
                         f'std_{rhs_name}': rhs_std,
                     }
-                t_stat, p_val = stats.ttest_ind(lhs_values, rhs_values)
-                stats_dict[metric][program]['t-value'] = t_stat
-                stats_dict[metric][program]['p-value'] = p_val
-                stats_dict[metric][program]['significant'] = "Y" if p_val < alpha else "N"
+                result = stats.ttest_ind(lhs_values, rhs_values)
```
tomershafir: Maybe a clearer name like `ttest` or `ttest_result`?

utils/compare.py
```diff
@@ -564,6 +582,13 @@ def main():
         default=False,
         help="Compute relative coefficient of variation (%%) rather than absolute stddev",
     )
+    parser.add_argument(
+        "--confidence-interval",
```
tomershafir: Respectively, I think `--diff-confidence-interval` might be clearer and also not too long

tomershafir (changes requested):

tomershafir (approved):
LGTM

PR #379, opened 2026-04-09
[LoopSplitting][MicroBenchmarks] Benchmarking `Loop Splitting` Transformation

loopacino:
PR adds benchmarking mechanics for the upstream PR on `#pragma omp split` directive


MicroBenchmarks/LoopSplit/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+# LoopSplit microbenchmark for #pragma omp split counts(...).
+# Copy this directory to llvm-test-suite/MicroBenchmarks/LoopSplit/
+# and add: add_subdirectory(LoopSplit) to MicroBenchmarks/CMakeLists.txt.
+#
+# Configure test-suite with a Clang that supports -fopenmp and -fopenmp-version=60.
```
Meinersbur, line 5:
```suggestion
```
remove instructions from AI that you obviously applied.

MicroBenchmarks/LoopSplit/main.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include <iostream>
+
+#include "benchmark/benchmark.h"
+
+// Large N for measurable runs; lit verification uses same kernel once.
+#define N 200000000
+
+// Kernel: sum 0..(N-1) with split into four segments.
+static long run_split() {
+  long sum = 0;
```
Meinersbur, line 13: long is 32 bits on 32 bit platforms and Windows 64. It will overflow on 19999999900000000

MicroBenchmarks/LoopSplit/main.cpp
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  benchmark::RunSpecifiedBenchmarks();
+  return EXIT_SUCCESS;
+}
+
+static void BM_Split(benchmark::State &state) {
+  long x = 0;
+  for (auto _ : state)
+    benchmark::DoNotOptimize(x += run_split());
```
Meinersbur, line 48:
```suggestion 
    auto x = run_split(); 
    benchmark::DoNotOptimize(x);
```

Meinersbur (commented):
https://llvm.org/docs/AIToolPolicy.html

> Contributors are expected to be transparent and label contributions that contain substantial amounts of tool-generated content. 

Benchmarks in llvm-test-suite are to track improvements of compiler optimizations. With `#pragma omp split` you already told the compiler what optimization to apply. There is very limited amout of optimization a compiler can do after that, basically applying the same optimization 4 times (each generated loop individually). There is no expectation it would improve in future versions of Clang. You are basically illustrating what speed difference of `#pragma omp split` for programmers considering using it, but it is a waste of time for compiler engineers that want to improve code optimization passes. That is, I don't think we need a benchmark for `split` upstream.

amitamd7:
> https://llvm.org/docs/AIToolPolicy.html
> 
> > Contributors are expected to be transparent and label contributions that contain substantial amounts of tool-generated content.
> 
> Benchmarks in llvm-test-suite are to track improvements of compiler optimizations. With `#pragma omp split` you already told the compiler what optimization to apply. There is very limited amout of optimization a compiler can do after that, basically applying the same optimization 4 times (each generated loop individually). There is no expectation it would improve in future versions of Clang. You are basically illustrating what speed difference of `#pragma omp split` for programmers considering using it, but it is a waste of time for compiler engineers that want to improve code optimization passes. That is, I don't think we need a benchmark for `split` upstream.

Understood. Thanks for reviewing and sharing the doc, will go through it.

MicroBenchmarks/LoopSplit/CMakeLists.txt
loopacino: Thanks for pointing that one out, will keep a note for future references.

PR #380, opened 2026-04-10
[CMake] Add `TEST_SUITE_SAMPLE_PROFILE_USE` for sample-based PGO

mustartt:
Add support for sample profile support (`-fprofile-sample-use`) to the test suite with two new CMake flags:

  - `TEST_SUITE_SAMPLE_PROFILE_USE=On|Off`: enables sample profile use.
  - `TEST_SUITE_SAMPLE_PROFILE_DIR=<path>`: a directory containing `<target>.spgo` sample profile files. This provides an easy harness to test-suite to guard against sample profile regressions.

Unlike instrumentation PGO which requires a two-phase generate + use workflow, sample profiles are collected externally, and the collection method various widely between platforms. For each benchmark target, the build system looks for `<dir>/<target>.spgo` and passes `-fprofile-sample-use=<path>` to the compile flags. In case, the target does not have an associated sample profile, a warning is emitted. In addition, when the sample profile changes, the target will also be recompiled.

```bash
# collection
cmake -DTEST_SUITE_BENCHMARKING_ONLY=On \
      -DTEST_SUITE_RUN_UNDER=<collection tool> \
      ../llvm-test-suite
make 
llvm-lit .

# merge sample profiles into /path/to/profiles
/path/to/profiles/ contains:
401.bzip2.spgo
429.mcf.spgo
456.hmmer.spgo
470.lbm.spgo
...

# final build
cmake -DTEST_SUITE_BENCHMARKING_ONLY=On \
      -DTEST_SUITE_SAMPLE_PROFILE_USE=On \
      -DTEST_SUITE_SAMPLE_PROFILE_DIR=/path/to/profiles \
      ../llvm-test-suite
...

### CODE REVIEW

**Comment by mustartt** [comment]
Gentle ping

File: cmake/modules/TestSuite.cmake [hunk 1]
```diff
@@ -58,6 +58,15 @@ function(llvm_test_executable_no_test target)
     append_target_flags(COMPILE_FLAGS ${target} -fprofile-instr-use=${target_path}.profdata)
     append_target_flags(LINK_LIBRARIES ${target} -fprofile-instr-use=${target_path}.profdata)
   endif()
+  if(TEST_SUITE_SAMPLE_PROFILE_USE)
+    set(spgo_file "${TEST_SUITE_SAMPLE_PROFILE_DIR}/${target}.spgo")
+    if(EXISTS "${spgo_file}")
+      append_target_flags(COMPILE_FLAGS ${target} -fprofile-sample-use=${spgo_file})
+      append_target_flags(LINK_LIBRARIES ${target} -fprofile-sample-use=${spgo_file})
+    else()
```
tobias-stadler, line 70: It would be nice if cmake recompiles when the profile changes.
Something like this maybe?
```suggestion
      get_target_property(_spgo_srcs ${target} SOURCES)
      foreach(_src ${_spgo_srcs})
        set_property(SOURCE ${_src} APPEND PROPERTY OBJECT_DEPENDS "${spgo_file}")
      endforeach()
    else()
```

mustartt: Yes, the same should be applied to https://github.com/llvm/llvm-project/blob/8f45e083ebe0e471d315ad1a4187730081c3dbdc/llvm/cmake/modules/HandleLLVMOptions.cmake#L1293

cmake/modules/TestSuite.cmake
```diff
@@ -58,6 +58,19 @@ function(llvm_test_executable_no_test target)
     append_target_flags(COMPILE_FLAGS ${target} -fprofile-instr-use=${target_path}.profdata)
     append_target_flags(LINK_LIBRARIES ${target} -fprofile-instr-use=${target_path}.profdata)
   endif()
+  if(TEST_SUITE_SAMPLE_PROFILE_USE)
+    set(spgo_file "${TEST_SUITE_SAMPLE_PROFILE_DIR}/${target}.spgo")
+    if(EXISTS "${spgo_file}")
+      append_target_flags(COMPILE_FLAGS ${target} -fprofile-sample-use=${spgo_file})
+      append_target_flags(LINK_LIBRARIES ${target} -fprofile-sample-use=${spgo_file})
```
apolloww, line 65: I wonder if the test suite supports ThinLTO. If so, then extra linker flag `--lto-sample-profile` or `--plugin-opt=sample-profile` is needed.

mustartt: Hi, thanks for taking a look.

Yes, test-suite does support ThinLTO. For the majority of platforms `tools::addLTOOptions` https://github.com/llvm/llvm-project/blob/a88714bd5557a5df51c80f2748bea4390e0f96d5/clang/lib/Driver/ToolChains/CommonArgs.cpp#L1967
Will automatically forward the appropriate flag for the supplied linker flavor. 

This is actually one of the motivating reason for this patch. Our Darwin driver doesn't use the default `tools::addLTOOptions`, and thus missed the sample profile flag when constructing the link job (I will put up a patch for that later).

apolloww (approved):
LGTM

mustartt:
Thanks for taking a look!

Rebased on top of main to resolve the merge conflict.

PR #381, opened 2026-04-12
[SingleSource] Avoid race condition in symlink creation for forced opt vectorizer tests

asb:
Commit 6ac44557d81c7e2cf6ead140e7b6325cbc3d5c74 added variants of the vectorizer tests, built from the same source file. But each variant uses the same reference output (e.g. `early-exit.reference_output`). A symlink to this file is made in the build directory, and for parallel builds we can get an error like:

```
CMake Error: failed to create symbolic link '/home/buildbot-worker/bbroot/clang-riscv-gauntlet/llvm-test-suite/build.rva23/SingleSource/UnitTests/Vectorizer/early-exit.reference_output': File exists
```

Essentially the same problem was fixed before in-tree in a086e609ae7be23c1b4b86e79a7e25d9bf278085. This patch implements a fix along the same lines. llvm_singlesource now accepts a `DEST_SUFFIX` which is plumbed through so that a unique symlink is created per test variant.

A script like the following will reproduce the issue:

```
targets="
SingleSource/UnitTests/Vectorizer/early-exit
SingleSource/UnitTests/Vectorizer/tf-early-exit
SingleSource/UnitTests/Vectorizer/tfactivelanemask-early-exit
SingleSource/UnitTests/Vectorizer/ftic-early-exit
SingleSource/UnitTests/Vectorizer/tf-ftic-early-exit
SingleSource/UnitTests/Vectorizer/tfactivelanemask-ftic-early-exit
"

for i in $(seq 1 50); do
  echo "attempt $i"

  rm -f SingleSource/UnitTests/Vectorizer/early-exit.reference_output
  ninja -t clean $targets >/dev/null || exit 1
  if ! ninja -j32 $targets; then
    echo "reproduced: parallel build failed; look above for the create_symlink File exists error"
    exit 0
  fi
done

echo "did not reproduce"
exit 1
```

I made use of an LLM (GPT 5.4) to debug the failure, find an in-tree example of the same issue being worked around, and to apply that pattern here.

---

This issue is causing spurious failures on the RISC-V buildbots right now (I would expect others too?). So it would be good to find a solution ASAP.

I get the failure with cmake 4.3.1 and ninja 1.13.2. I don't know if it's something that depends on cmake version.


fhahn (approved):
LGTM, thanks

PR #382, opened 2026-04-14
[cmake][Darwin] Add `TEST_SUITE_EXTERNALIZE_DEBUGINFO` for dSYM generation.

mustartt:
Add a `TEST_SUITE_EXTERNALIZE_DEBUGINFO` CMake option that runs `dsymutil` as a post build step on Darwin to generate .dSYM bundles for each target on Darwin. When `-flto` flags are detected, this passes `-Wl,-object_path_lto` so `dsymutil` still has access to the original objects through OSO table.

### DIFF
--- a/cmake/modules/TestSuite.cmake
+++ b/cmake/modules/TestSuite.cmake
@@ -64,16 +64,21 @@ function(llvm_test_executable_no_test target)
   test_suite_add_build_dependencies(${target})
 
   if(TEST_SUITE_EXTERNALIZE_DEBUGINFO AND APPLE)
+    string(TOUPPER "${CMAKE_BUILD_TYPE}" uppercase_CMAKE_BUILD_TYPE)
     if(CMAKE_C_FLAGS MATCHES "-flto"
       OR CMAKE_CXX_FLAGS MATCHES "-flto"
-      OR CMAKE_C_FLAGS_RELEASE MATCHES "-flto"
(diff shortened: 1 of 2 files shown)


cmake/modules/TestSuite.cmake
```diff
@@ -63,6 +63,22 @@ function(llvm_test_executable_no_test target)
...
+    if(CMAKE_C_FLAGS MATCHES "-flto"
+      OR CMAKE_CXX_FLAGS MATCHES "-flto"
+      OR CMAKE_C_FLAGS_RELEASE MATCHES "-flto"
+      OR CMAKE_CXX_FLAGS_RELEASE MATCHES "-flto")
+      set(lto_object ${CMAKE_CURRENT_BINARY_DIR}/${target}-lto.o)
+      set_property(TARGET ${target} APPEND_STRING PROPERTY
+        LINK_FLAGS " -Wl,-object_path_lto,${lto_object}")
+    endif()
+    if(NOT CMAKE_DSYMUTIL)
+      set(CMAKE_DSYMUTIL xcrun dsymutil)
```
dtellenbach: How about `xcrun -f dsymutil`?

dtellenbach (approved):
One nit, otherwise LGTM.

cmake/modules/TestSuite.cmake
mustartt: Sounds good. Used `xcrun -f dsymutil`'s to seed CMAKE_DSYMUTIL.

PR #383, opened 2026-04-15
[Flang][OpenMP] Fix depend_iterator regression (omp_lib.mod not found)

chichunchen:
Fix regression due to https://github.com/llvm/llvm-test-suite/pull/375 by adding OpenMP check in cmake and replace `use omp_lib` with external declarations. (avoid .mod dependency file but probably not needed with the cmake guard)

### DIFF
--- a/Fortran/cray/OpenMP/5.0/depend_iterator_02.f90
+++ b/Fortran/cray/OpenMP/5.0/depend_iterator_02.f90
@@ -6,8 +6,8 @@
 ! OpenMP implementations, but it is expected to pass for CCE and flang.
 
 program main
-  use omp_lib
   implicit none
+  external :: omp_set_dynamic, omp_set_num_threads
   integer :: lll,uuu,sss,aaa(2)
 
   lll = 1
--- a/Fortran/cray/OpenMP/CMakeLists.txt
+++ b/Fortran/cray/OpenMP/CMakeLists.txt
@@ -1,1 +1,3 @@
-add_subdirectory(5.0)
+if(OPENMP_FOUND)
+  add_subdirectory(5.0)
+endif()


ceseo (approved):
LGTM

tarunprabhu (approved):
Thanks for the quick fix. LGTM.

For my future reference, do you know where `OPENMP_FOUND` is set?

chichunchen:
> Thanks for the quick fix. LGTM.
> 
> For my future reference, do you know where `OPENMP_FOUND` is set?

https://github.com/llvm/llvm-test-suite/blob/3c51b05006edf8e2c9a53667b0acd551ac6d0917/CMakeLists.txt#L251
https://cmake.org/cmake/help/latest/module/FindOpenMP.html#result-variables

tarunprabhu:
> > Thanks for the quick fix. LGTM.
> > For my future reference, do you know where `OPENMP_FOUND` is set?
> 
> https://github.com/llvm/llvm-test-suite/blob/3c51b05006edf8e2c9a53667b0acd551ac6d0917/CMakeLists.txt#L251
> 
> https://cmake.org/cmake/help/latest/module/FindOpenMP.html#result-variables

In this case, would it be safer to specifically check for `OPEN_Fortran_FOUND`? Will `OPENMP_FOUND` return true even if `omp.mod` was not found?

ceseo:
> > > Thanks for the quick fix. LGTM.
> > > For my future reference, do you know where `OPENMP_FOUND` is set?
> > 
> > 
> > https://github.com/llvm/llvm-test-suite/blob/3c51b05006edf8e2c9a53667b0acd551ac6d0917/CMakeLists.txt#L251
> > 
> > https://cmake.org/cmake/help/latest/module/FindOpenMP.html#result-variables
> 
> In this case, would it be safer to specifically check for `OPEN_Fortran_FOUND`? Will `OPENMP_FOUND` return true even if `omp.mod` was not found?

Is it possible to enable OpenMP for C++ but not Fortran in the LLVM build? If so, then yes, `OPENMP_Fortran_FOUND` is more adequate.

tarunprabhu:
> Is it possible to enable OpenMP for C++ but not Fortran in the LLVM build? 

That's a good point. I don't know. Let's see if this resolves the issues for now. It may be worth looking into just in case.

ceseo:
I think there is another problem.

The `depend_iterator_01/02` tests now show as `Executable Missing` instead of `FAIL`. This PR guards `add_subdirectory(5.0)` with `if(OPENMP_FOUND)` but LIT still discovers the `.test` files and flags them as missing executables. Probably needs a change that also suppresses LIT discovery.

PR #385, opened 2026-04-20
[Fortran/gfortran] Disabling few coarray tests after merged PR#182110

JDPailleux:
Disables tests on coarrays after PR  https://github.com/llvm/llvm-project/pull/182110  has been merged


jeanPerier (approved):
Thanks

tarunprabhu (approved):
Thanks. Just a couple of questions:

Do you know why these worked before? 

Do you think they can be re-enabled the future as more coarray support is added?

JDPailleux:
Thanks !
@tarunprabhu The allocation of a coarray was ignored in Flang, which might explain why the tests passed previously. 
Yes, I think the tests can be re-enabled later with few adjustments maybe.

PR #386, opened 2026-04-20
[NFC] Rename PreferPredicateOverEpilogue to TailFoldingPolicy

hassnaaHamdi:
The flag of -prefer-predicate-over-epilogue got changed in the LLVM source code into -tail-folding-policy.
Update the test suite accordingly.

### DIFF
--- a/SingleSource/UnitTests/Vectorizer/CMakeLists.txt
+++ b/SingleSource/UnitTests/Vectorizer/CMakeLists.txt
@@ -13,7 +13,7 @@ if ("${CMAKE_C_COMPILER_ID}" MATCHES "Clang")
 
   # Add targets for all tests with forced tail folding.
   list(APPEND CXXFLAGS
-    "-mllvm" "-prefer-predicate-over-epilogue=predicate-dont-vectorize"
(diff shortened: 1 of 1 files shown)


david-arm (approved):
LGTM!

SingleSource/UnitTests/Vectorizer/CMakeLists.txt
```diff
@@ -13,15 +13,15 @@ if ("${CMAKE_C_COMPILER_ID}" MATCHES "Clang")
 
   # Add targets for all tests with forced tail folding.
   list(APPEND CXXFLAGS
-    "-mllvm" "-prefer-predicate-over-epilogue=predicate-dont-vectorize"
+    "-mllvm" "-tail-folding-policy=must-fold-tail"
```
fhahn, line 16: Could we check if the new/old options are available, otherwise it may be painful to use/compare against slightly older version of Clang (e.g. CI/jobs running llvm-test-suite against a release branch)?

hassnaaHamdi: Okay 👍 thank you.

PR #387, opened 2026-04-22
[Fortran][test-suite] Update empty BIND(C) derived type expectations

Saieiei:
This updates the Fortran test-suite fallout from llvm/llvm-project#193452.

Flang now correctly diagnoses empty `BIND(C)` derived types as errors, so `empty_derived_type.f90` is marked `xfail`, and `empty_derived_type_2.f90` is removed from `FAILING_FILES` now that it produces the expected error.

### DIFF
--- a/Fortran/gfortran/regression/override.yaml
+++ b/Fortran/gfortran/regression/override.yaml
@@ -139,3 +139,7 @@
 # on big endian target.
 "unsigned_21.f90":
   disabled_on: ["powerpc-*-aix"]
+
+# Flang now diagnoses empty BIND(C) derived types as errors.
+"empty_derived_type.f90":
+  xfail: true
--- a/Fortran/gfortran/regression/DisabledFiles.cmake
+++ b/Fortran/gfortran/regression/DisabledFiles.cmake
@@ -1011,7 +1011,6 @@ file(GLOB FAILING_FILES CONFIGURE_DEPENDS
   allocate_error_8.f90
   do_concurrent_12.f90
   do_concurrent_15.f90
-  empty_derived_type_2.f90
   interface_51.f90
   interface_52.f90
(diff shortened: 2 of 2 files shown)


tarunprabhu (changes requested):
Please do not update tests.cmake by hand. There is a large banner at the top that says that file must not be edited. 

If the file is xfail in flang but is expected to pass in gfortran, see the [README.md](https://github.com/llvm/llvm-test-suite/blob/main/Fortran/gfortran/README.md#overriding-dejagnu-annotations) on how to use override.yaml to override the annotations.

Saieiei:
My bad, I should have caught that. Thanks.

I switched this to use `override.yaml`, regenerated `tests.cmake` with `utils/update-test-config.py`, and updated the PR. With the corrected changes, both `empty_derived_type.f90` and `empty_derived_type_2.f90` now pass.

Fortran/gfortran/regression/override.yaml
```diff
@@ -139,3 +139,7 @@
 # on big endian target.
 "unsigned_21.f90":
   disabled_on: ["powerpc-*-aix"]
+
+# Flang now diagnoses empty BIND(C) derived types as errors.
```
tarunprabhu, line 143: Do you mind providing a bit more context here? Something like the following. 

```suggestion
# Flang now diagnoses empty BIND(C) derived types as errors. gfortran issues
# a warning in this case.
```

tarunprabhu (commented):
Thanks. A small request and we should be good to go.

Saieiei:
Closing this follow-up too, since it depends on the hard-error direction in https://github.com/llvm/llvm-project/pull/193452 that’s no longer moving forward.

Thanks for the review and guidance on the llvm-test-suite workflow!!

PR #390, opened 2026-04-22
[SingleSource] Check FindLast tests with low trip counts.

fhahn:
Checking with low trip counts reproduced a now fixed tail-folding mis-compile on plain AArch64 NEON.


lukel97 (approved):

Mel-Chen (approved):
LG, but there is a duplicated "a" in the PR description. 
Should be "Checking with low trip counts reproduced a now fixed tail-folding mis-compile on plain AArch64 NEON."?

PR #391, opened 2026-04-23
[fpcmp] Fix crash when comparing an empty file

carlsotr:
This fixes a crash that can be reproduced as follows.

```
$ echo hello > a
$ echo -n > b
$ ./fpcmp a b
./fpcmp: Comparison failed, unexpected end of one of the files

Input 1:
hello
Segmentation fault (core dumped)
```

### DIFF
--- a/tools/fpcmp.c
+++ b/tools/fpcmp.c
@@ -212,12 +212,6 @@ char *load_file(const char *path, long *size_out) {
     exit(2);
   }
 
-  /* Avoid malloc(0). */
-  if (size == 0) {
-    *size_out = 0;
-    return 0;
-  }
-
   /* Allocate a buffer for the data. */
   data = malloc(size + 1);
   if (!data) {
@@ -228,7 +222,7 @@ char *load_file(const char *path, long *size_out) {
 
   /* Read in the file contents. */
   data[size] = 0;
-  if (fread(data, size, 1, fp) != 1) {
+  if (size != 0 && fread(data, size, 1, fp) != 1) {
     fprintf(stderr, "%s: error: unable to read data for '%s'\n",
             g_program, path);
     exit(2);


tools/fpcmp.c
```diff
@@ -249,7 +249,9 @@ static bool contains_non_printable_characters(const char *data) {
 }
 
 static void dump_input(const char *label, const char *data) {
-  if (contains_non_printable_characters(data)) {
+  if (data == 0) {
```
tarunprabhu: If the file is empty, is it possible for `data` to be an empty string?

tarunprabhu (commented):
Thanks for working on this.

nikic (approved):
LGTM

tools/fpcmp.c
carlsotr: Yes, I think that would be a better solution. How does this look? That "Avoid malloc(0)" comment appears to be stale since e3d31a7357112a0abde2bb677a8d8d5e3d810a9c changed `malloc(size)` to `malloc(size + 1)`. If you prefer this, I'll force push it.

```diff
--- a/tools/fpcmp.c
+++ b/tools/fpcmp.c
@@ -212,12 +212,6 @@ char *load_file(const char *path, long *size_out) {
     exit(2);
   }

-  /* Avoid malloc(0). */
-  if (size == 0) {
-    *size_out = 0;
-    return 0;
-  }
-
   /* Allocate a buffer for the data. */
   data = malloc(size + 1);
   if (!data) {
@@ -228,7 +222,7 @@ char *load_file(const char *path, long *size_out) {

   /* Read in the file contents. */
   data[size] = 0;
-  if (fread(data, size, 1, fp) != 1) {
+  if (size != 0 && fread(data, size, 1, fp) != 1) {
     fprintf(stderr, "%s: error: unable to read data for '%s'\n",
             g_program, path);
     exit(2);
```

Output:
```
$ ./fpcmp-391-new a b
./fpcmp-391-new: Comparison failed, unexpected end of one of the files

Input 1:
hello

Input 2:
$ ./fpcmp-391-orig a b
./fpcmp-391-orig: Comparison failed, unexpected end of one of the files

Input 1:
hello

Input 2: Empty file.
$

```

tarunprabhu: Thanks. This feels a bit cleaner since we are guaranteed that data is not `null`. We can also be certain that it will be an empty string.

tarunprabhu (approved):
Thanks for the changes. LGTM.

tools/fpcmp.c
carlsotr: Agreed. Thank you for the suggestion and review. I've pushed the change.

carlsotr:
I don't have permission to merge. @tarunprabhu, would you please merge this for me?

PR #392, opened 2026-04-23
[Flang][Windows] Don't try to run Unix-specific tests on Windows

DavidTruby:
Some of these recently added tests call functions only available in POSIX, and also assume that the linker accepts -fopenmp. These won't work on Windows, so we shouldn't attempt to build or run them there.


chichunchen (approved):
Thanks for the fix.

tarunprabhu (approved):
Thanks, David.

PR #394, opened 2026-04-23
[UnitTest] Check new tail-folding flag, if it fails, fall back to old one

hassnaaHamdi:


hassnaaHamdi:
hello @fhahn :'D

fhahn (approved):
LGTM, thanks for fixing this.

Would be good to update the title to include something like `[UnitTest]` and capitalize first letter.

hassnaaHamdi:
> LGTM, thanks for fixing this.
> 
> Would be good to update the title to include something like `[UnitTest]` and capitalize first letter.

Thank you :))

PR #396, opened 2026-04-27
[flang] added write-after-endfile testcase

blazie2004:
This test verifies that Fortran I/O statements correctly handle error conditions when ERR= and IOSTAT= specifiers are present. Per Fortran standard 12.11.2, when an error occurs during I/O and either ERR= or IOSTAT= is specified.

PR: [191633](https://github.com/llvm/llvm-project/pull/191633)


tarunprabhu:
@kwyatt-ext, 

Is this test being added to the correct subdirectory of the cray tests?

kwyatt-ext:
I will review this.  I'm contacting the author about the related PR first.

Fortran/cray/f90/10/write-after-endfile.f90
```diff
@@ -0,0 +1,52 @@
+program test_io_err
```
kwyatt-ext: The Fortran '90 Standard has the Input/Output defined under Section 9.  I'm not sure about sub-section.  This file would probably be stored under `./f90/09/xx`.  Since this is a test re-write, it might make more sense to use a more modern standard.  I believe I saw a comment citing the requirement this demonstrates as being F2023 12.11.2.  If that is correct, then the location of this test should be `./f2023/12/11`.

kwyatt-ext: The program name here and at the end should match the test name.  This is the safest method of naming, depending upon how the test system runs them.

Fortran/cray/f90/10/write-after-endfile.f90
```diff
@@ -0,0 +1,52 @@
...
+  print *, 'iostat after write:', n, ' k:', k
+  ok = ok .and. (k == 66)
+
+  if (ok) then
+    print *, 'PASS'
+  else
+    print *, 'FAIL'
+  end if
+
+end program test_io_err
```
kwyatt-ext: A couple of overall issues: 
- You are missing the corresponding reference_output file.
- As is, it currently "passes" the llvm test suite.  However, if I run this test manually, it core dumps.  You will need to wait for the corresponding code fix in order to merge this (after the review is complete).

Fortran/cray/f90/10/write-after-endfile.f90
```diff
@@ -0,0 +1,52 @@
+program test_io_err
+
+  implicit none
+  integer :: n, k, l
+  logical :: ok
+
+  ok = .true.
+
+  open(5, err=500, iostat=n)
```
kwyatt-ext: Just to clarify, as I've not much experience with Fortran I/O, These first two cases are supposed to be valid and NOT jump to their error condition, but the second two cases ARE?

kwyatt-ext (changes requested):

Fortran/cray/f90/10/write-after-endfile.f90
kwyatt-ext: Also, the name "write-after-endfile" isn't a little odd.  You are writing after a file was closed.  I'm not sure if that is the same as writing past the end of a file.

Fortran/cray/f90/10/write-after-endfile.f90
kwyatt-ext: I researched it.  The OPEN/CLOSE should pass as is.  The READ/WRITE on a closed file should not.

kwyatt-ext:
Hi @blazie2004.  I have a couple of general recommendations:

- Please don't force-push successive commits to a PR.  It erases the history.  If a reviewer can't see the diff between their comments and your current push, they end up reviewing the whole change each time.  The extra commit history will be compressed into one entry when the change is squashed and merged.

- Please hit the "Re-request Review", button when you are ready for the reviewer to look again.  I assume you are ready since you did a force-push, but am not sure if you intended to hand it back at this point.

Thank you.

tarunprabhu:
@blazie2004, Please see LLVM's developer guide on [pull requests](https://llvm.org/docs/GitHub.html#pull-requests) - in particular, the sections on updating and rebasing PR's.

Fortran/cray/f90/10/write-after-endfile.f90
blazie2004: Added the testcase here : Fortran/cray/f2023/12/11/io-after-close.f90

blazie2004: resolved

Fortran/cray/f90/10/write-after-endfile.f90
blazie2004: resolved

Fortran/cray/f90/10/write-after-endfile.f90
blazie2004: changed it to :  io-after-close.f90

kwyatt-ext: You need to delete write-after-endfile.f90 from your commit.

kwyatt-ext (changes requested):
After you remove the extra file, this should be ready to go, with the following exceptions:
- It still will NOT pass with the draft PR code changes.
- It can not be merged until the associated fix has been merged.

Fortran/cray/f2023/12/11/io-after-close.reference_output
```diff
@@ -0,0 +1,5 @@
+k after OPEN:          45
+k after CLOSE:          55
+iostat after read:           0  k:          65
+iostat after write:           0  k:          65
+FAIL
```
eugeneepshteyn: The test fails?

kwyatt-ext: Eugene is right.  This expected output is not correct.  If your test is successful in branching to the error label for the last 2 cases, then your "k" value should be "66" for both and the final should be PASS.

Fortran/cray/f2023/12/11/io-after-close.reference_output
```diff
@@ -0,0 +1,5 @@
+k after OPEN:          45
+k after CLOSE:          55
+iostat after read:         125  k:          66
+iostat after write:         125  k:          66
+PASS
```
eugeneepshteyn: github shows missing end-of-line before the end of file

Fortran/cray/f2023/12/11/io-after-close.f90
```diff
@@ -0,0 +1,53 @@
+program io_after_close
+
+  implicit none
+  integer :: n, k, l
+  logical :: ok
+
+  ok = .true.
+
```
eugeneepshteyn: Note that flang-rt PR is a fix in `BeginUnformattedIO()`, yet there's no `open` statement here with `form='unformatted'`. Please add tests with `unformatted`.

Fortran/cray/f2023/12/11/io-after-close.f90
```diff
@@ -0,0 +1,53 @@
...
+  open(5, err=500, iostat=n)
```
eugeneepshteyn: Note that unit 5 is often a pre-connected unit for standard input. I highly recommend you change the number to something else, or use `newUnit`.

eugeneepshteyn (changes requested):

Fortran/cray/f2023/12/11/io-after-close.f90
blazie2004: resolved

Fortran/cray/f2023/12/11/io-after-close.f90
blazie2004: resolved

Fortran/cray/f2023/12/11/io-after-close.reference_output
blazie2004: resolved

kwyatt-ext:
@blazie2004

Please try this test against your fix.  It will FAIL.

In Flang, the READ of a closed file is treated as an EOF condition, not an error condition.  This is acceptable, as the behavior is "processor dependent".  However, your test for the closed READ case is still testing for an END error and will fail.  The expected results _should_ be:
`iostat after read: -1  k: 65`

Or, you could change the closed READ case to an `EOF=` instead of `ERR=`.

Fortran/cray/f2023/12/11/io-after-close.f90
```diff
@@ -0,0 +1,51 @@
...
+   goto 515
+510 continue
+   k = 56
+515 continue
+  print *, 'k after CLOSE:', k
+  ok = ok .and. (k == 55)
+
+  ! Flang may treat READ from a closed unit as EOF or as another
+  ! processor-dependent condition. Do not check this result here.
+  read(iounit, end=520, iostat=n) l
```
kwyatt-ext: This is just a question, not a required change.  I'm curious why you left this read in, since the test portion has been removed.  Did it cause the next write to behave incorrectly?

Fortran/cray/f2023/12/11/io-after-close.f90
```diff
@@ -0,0 +1,51 @@
...
+  print *, 'iostat after write:', n, ' k:', k
+  ok = ok .and. (k == 66)
+
+  if (ok) then
+    print *, 'PASS'
+  else
+    print *, 'FAIL'
+  end if
+
+end program io_after_close
```
kwyatt-ext: No newline at end of file

kwyatt-ext (changes requested):
Once the newline has been added, it will LGTM.  I am curious about the question I asked in this review though.

Fortran/cray/f2023/12/11/io-after-close.f90
```diff
@@ -0,0 +1,51 @@
...
+   goto 505
+500 continue
+   k = 46
+505 continue
+  print *, 'k after OPEN:', k
+  ok = ok .and. (k == 45)
+
+  write(iounit) 123
+
+  close(iounit, err=510, iostat=n)
```
eugeneepshteyn: Description of the implementation PR says "A WRITE after ENDFILE ...". This test is for WRITE after CLOSE. Please change the test to correspond to the PR.

Fortran/cray/f2023/12/11/io-after-close.f90
blazie2004: No, the leftover READ did not affect the following WRITE. I checked this on my build: the READ on the closed unit returned `iostat=1042` (`IostatBadUnitNumber`), and the WRITE after that also returned `1042`, because both were happening on a closed unit.

Because of that, neither of them went through the write-after-endfile path that this PR is fixing. That path should return `iostat=1017` (`IostatWriteAfterEndfile`).

So the READ was just leftover/unnecessary code. I removed it completely and changed the test to focus on the actual fix. I also added the missing trailing newline.

Fortran/cray/f2023/12/11/io-after-close.f90
blazie2004: resolved

Fortran/cray/f2023/12/11/io-after-close.f90
blazie2004: resolved

eugeneepshteyn (approved):
LGTM, thank you

kwyatt-ext (approved):

cenewcombe:
@blazie2004 @kwyatt-ext I believe this commit is causing buildbot failures: https://lab.llvm.org/buildbot/#/builders/121/builds/2433

antmox:
hi @kwyatt-ext, hi @blazie2004 
some additional failures since this patch was merged:
https://lab.llvm.org/buildbot/#/builders/222/builds/9784
https://lab.llvm.org/buildbot/#/builders/17/builds/15967
https://lab.llvm.org/buildbot/#/builders/41/builds/12138
https://lab.llvm.org/buildbot/#/builders/143/builds/15680
https://lab.llvm.org/buildbot/#/builders/4/builds/12353
https://lab.llvm.org/buildbot/#/builders/198/builds/13750
https://lab.llvm.org/buildbot/#/builders/199/builds/9983

kwyatt-ext:
Looking now.  Should only be a quick fix.  It looks like the results file might be messed up.

kwyatt-ext:
Reverted to allow the author to correct it when available.

eugeneepshteyn:
@kwyatt-ext , note that with llvm-test-suite tests, while reverting may be easier, another perfectly valid course of action is to simply disable the test.

eugeneepshteyn:
@blazie2004 , I think the status code may be system dependent, so it's best to avoid exact status code values in the result file.

PR #398, opened 2026-05-01
[Apple] Support benchmarking multiple Mach-O slices

jroelofs:
This is enabled through a new CMake variable `TEST_SUITE_RUN_ARCHS` that allows users to specify which subset arches from CMAKE_OSX_ARCHITECTURES they'd like to run the tests for. By default, when there's two or more, we emit foo.<slice>.test files for all of them. Existing names are preserved when TEST_SUITE_RUN_ARCHS is a list of size zero or one.

Implementation-wise, we extend the `timeit` binary to support `-arch <slice>`, much like macOS's `arch` utility, and plumb in the requested slice name via a new `ARCH:` directive in the .test files.

### DIFF
--- a/tools/macho_arch.c
+++ b/tools/macho_arch.c
@@ -7,22 +7,12 @@
 |*                                                                            *|
 \*===----------------------------------------------------------------------===*/
 
-#include <errno.h>
 #include <mach-o/ldsyms.h>
 #include <mach-o/utils.h>
-#include <mach/machine.h>
 #include <stdio.h>
-#include <string.h>
-#include <sys/sysctl.h>
 
 int main(int argc, const char **argv) {
   if (__builtin_available(macOS 13.0, iOS 16.0, tvOS 16.0, watchOS 9.0, *)) {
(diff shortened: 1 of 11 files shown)


tools/macho_arch.c
```diff
@@ -0,0 +1,36 @@
...
+#include <stdio.h>
+#include <string.h>
+#include <sys/sysctl.h>
+
+int main(int argc, const char **argv) {
+  if (__builtin_available(macOS 13.0, iOS 16.0, tvOS 16.0, watchOS 9.0, *)) {
+    if (argc > 2 && strcmp(argv[1], "--is-supported") == 0) {
+      cpu_type_t cpu_type;
+      cpu_subtype_t cpu_subtype;
+      return macho_cpu_type_for_arch_name(argv[2], &cpu_type, &cpu_subtype) ? 0 : 1;
```
jroelofs: FIXME: this isn't the right API for "does $host support $slice"... it's barely a weak proxy for "does $host's dyld know about $slice"

aemerson (approved):

PR #399, opened 2026-05-11
[tools] Tolerate test_maxrss failures on Linux

tomershafir:
`tools/test/test_maxrss.py` sometimes fails on Linux build bots. See: https://github.com/llvm/llvm-test-suite/pull/373

The max RSS number reported by the OS is generally not strictly reliable and can have exhibit non-deterministic behavior. Thus, this patch only checks conservatively that the number is just greater than 0.

### DIFF
--- a/tools/test/test_maxrss.py
+++ b/tools/test/test_maxrss.py
@@ -36,9 +36,12 @@ try:
         print("maxrss line not found in summary:\n%s" % contents)
         sys.exit(1)
 
-    expected_min = ALLOC_MiB * 1024 * 1024
-    if maxrss < expected_min:
-        print("maxrss too low: %d < %d" % (maxrss, expected_min))
+    # The absolute value of maxrss reported by the OS is not strictly reliable,
+    # thus we conservatively just check we could capture any number > 0, and not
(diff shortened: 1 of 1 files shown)


tools/test/test_maxrss.py
```diff
@@ -37,6 +37,9 @@
         sys.exit(1)
 
     expected_min = ALLOC_MiB * 1024 * 1024
+    # Tolerate a 10% failure to accommodate (seemingly deterministic) flakiness on Linux
+    # See: https://github.com/llvm/llvm-test-suite/pull/373
+    expected_min = int(expected_min * 0.9)
```
lukel97: My recommendation would still be to just drop this check since there's nothing that guarantees a sample will be taken between the malloc and the free call. And AFAIK the kernel doesn't even have to return an accurate ru_maxrss to begin with. So I wouldn't worry about having to test the value at all

tomershafir: re-thinking yea I think you are right, fixing

lukel97 (approved):
LGTM, thanks for fixing

PR #404, opened 2026-05-14
Add MicroBenchmark for Small Trip Count Loop vectorization

Stylie777:
For targets where getMinTripCountTailFoldingThreshold returns a value greater than zero, https://github.com/llvm/llvm-project/pull/195823 has enabled better vectorization of loops where applicable. This micro benchmark is intended to show the impact of these changes on the relevant targets.

For targets where getMinTripCountTailFoldingThreshold returns zero, there will be no effect to runtime when comparing scalar vs vector.

Assisted-by: Codex

### DIFF
--- a/MicroBenchmarks/LoopVectorization/SmallLoopTripCount.cpp
+++ b/MicroBenchmarks/LoopVectorization/SmallLoopTripCount.cpp
@@ -1,0 +1,102 @@
+// This program tests the performance impact of vectorization in loops with
+// small trip counts. These cases exercise the LoopVectorize path that accepts
+// trip counts one larger than the vectorization factor.
+
+#include <array>
+#include <cstddef>
+#include <cstdint>
+#include <limits>
+
+#include "benchmark/benchmark.h"
+
+#define NOINLINE __attribute__((noinline))
+#define LOOP_VECTORIZE_ENABLE                                                  \
+  _Pragma("clang loop vectorize(enable) unroll(disable)")
+#define LOOP_VECTORIZE_DISABLE                                                 \
+  _Pragma("clang loop vectorize(disable) interleave(disable) unroll(disable)")
+#define LOOP_INTERLEAVE_COUNT_2                                                \
+  _Pragma("clang loop vectorize(enable) interleave_count(2) unroll(disable)")
+
+static uint64_t g_small_loop_trip_count_sum = 0;
+
+template <typename Ty>
+NOINLINE void loopTc5Vector(const Ty *__restrict A, Ty *__restrict B) {
+  LOOP_VECTORIZE_ENABLE
+  for (uint64_t I = 0; I != 5; ++I)
+    B[I] = A[I] + static_cast<Ty>(1);
+}
+
+template <typename Ty>
+NOINLINE void loopTc5Scalar(const Ty *__restrict A, Ty *__restrict B) {
+  LOOP_VECTORIZE_DISABLE
+  for (uint64_t I = 0; I != 5; ++I)
+    B[I] = A[I] + static_cast<Ty>(1);
+}
+
+NOINLINE void loopTc5I64InterleaveCount2Vector(const uint64_t *__restrict A,
+                                               uint64_t *__restrict B) {
+  LOOP_INTERLEAVE_COUNT_2
+  for (uint64_t I = 0; I != 5; ++I)
+    B[I] = A[I] + 1;
+}
+
+template <typename Ty> using KernelFn = void (*)(const Ty *, Ty *);
+
+template <typename Ty> static void initData(std::array<Ty, 16> &A) {
+  for (size_t I = 0; I != A.size(); ++I)
+    A[I] = static_cast<Ty>(0x0102030405060708ULL + I);
+}
+
+template <typename Ty> static uint64_t checksum(const std::array<Ty, 16> &A) {
+  uint64_t Sum = 0;
+  for (size_t I = 0; I != 5; ++I) {
+    auto Value = static_cast<uint64_t>(A[I]);
+    for (size_t Byte = 0; Byte != sizeof(Ty); ++Byte) {
+      Sum = Sum * 131 + (Value & std::numeric_limits<uint8_t>::max());
+      Value >>= 8;
+    }
+  }
+  return Sum;
+}
+
+template <typename Ty>
+static void runBenchForSmallLoopTripCount(benchmark::State &State,
+                                          KernelFn<Ty> Fn) {
+  std::array<Ty, 16> A;
+  std::array<Ty, 16> B = {};
+  initData(A);
+
+  for (auto _ : State) {
+    benchmark::DoNotOptimize(A.data());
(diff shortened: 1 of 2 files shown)


MicroBenchmarks/LoopVectorization/SmallLoopTripCount.cpp
```diff
@@ -0,0 +1,102 @@
...
+  for (auto _ : State) {
+    benchmark::DoNotOptimize(A.data());
+    benchmark::DoNotOptimize(B.data());
+    Fn(A.data(), B.data());
+    benchmark::ClobberMemory();
+  }
+
+  g_small_loop_trip_count_sum ^= checksum(B);
+  benchmark::DoNotOptimize(g_small_loop_trip_count_sum);
+  State.SetItemsProcessed(State.iterations() * 5);
```
fhahn: Would be good to comment why this is needed

MicroBenchmarks/LoopVectorization/SmallLoopTripCount.cpp
```diff
@@ -0,0 +1,102 @@
...
+}
+
+template <typename Ty>
+NOINLINE void loopTc5Scalar(const Ty *__restrict A, Ty *__restrict B) {
+  LOOP_VECTORIZE_DISABLE
+  for (uint64_t I = 0; I != 5; ++I)
+    B[I] = A[I] + static_cast<Ty>(1);
+}
+
+NOINLINE void loopTc5I64InterleaveCount2Vector(const uint64_t *__restrict A,
```
fhahn: Is there a reason to not use the templated version for this one as well?

MicroBenchmarks/LoopVectorization/SmallLoopTripCount.cpp
```diff
@@ -0,0 +1,102 @@
...
+                                          loopTc5I64InterleaveCount2Vector);
+}
+
+BENCHMARK_TEMPLATE(benchTc5Vector, uint8_t)->Name("tc5/i8/vector");
+BENCHMARK_TEMPLATE(benchTc5Scalar, uint8_t)->Name("tc5/i8/scalar");
+BENCHMARK_TEMPLATE(benchTc5Vector, uint16_t)->Name("tc5/i16/vector");
+BENCHMARK_TEMPLATE(benchTc5Scalar, uint16_t)->Name("tc5/i16/scalar");
+BENCHMARK_TEMPLATE(benchTc5Vector, uint32_t)->Name("tc5/i32/vector");
+BENCHMARK_TEMPLATE(benchTc5Scalar, uint32_t)->Name("tc5/i32/scalar");
+BENCHMARK_TEMPLATE(benchTc5Vector, uint64_t)->Name("tc5/i64/vector");
```
fhahn, line 281: I think the potential worst case would be i64 with TC =3, could you also cover this?

MicroBenchmarks/LoopVectorization/SmallLoopTripCount.cpp
```diff
@@ -0,0 +1,102 @@
...
+#define LOOP_INTERLEAVE_COUNT_2                                                \
+  _Pragma("clang loop vectorize(enable) interleave_count(2) unroll(disable)")
+
+static uint64_t g_small_loop_trip_count_sum = 0;
+
+template <typename Ty>
+NOINLINE void loopTc5Vector(const Ty *__restrict A, Ty *__restrict B) {
+  LOOP_VECTORIZE_ENABLE
+  for (uint64_t I = 0; I != 5; ++I)
+    B[I] = A[I] + static_cast<Ty>(1);
```
fhahn, line 23: This is a case where there is basically no overhead for the vector code compared to the scalar code.

Would be good to also include cases where there is some overhead from the vector code  compared to scalar, e.g. some scalarization

Stylie777: I have added an example that has scalarization in the loop, if this is not what you meant please let me know!

MicroBenchmarks/LoopVectorization/SmallLoopTripCount.cpp
Stylie777: I have added cases for all data types for TC=3 for full coverage.

MicroBenchmarks/LoopVectorization/SmallLoopTripCount.cpp
Stylie777: No there isn't, I have made it consistent now.

MicroBenchmarks/LoopVectorization/SmallLoopTripCount.cpp
Stylie777: It's not. I missed this when first reviewing the codex generated benchmark. I've removed it.

Stylie777:
Gentle Ping

fhahn (commented):
Could you add some of the following cases and double-check performance? On the system I tried (Apple M4 with `getMinTripCountTailFoldingThreshold` updated to return 5 even without SVE), the scalar code is noticeably faster than the vector code


with `int8_t` and `int16_t`
```
  for (int i = 0; i < 3; i++) { a[i] = a[i] < b[i] ? a[i] : b[i]; }
```

```
  for (int i = 0; i < 5; i++) { a[i]^=b[i]; }
```

with `int32_t` 
```
  for (int i = 0; i < 5; i++) { a[i] %= (b[i] | 1); }
```

Stylie777:
@fhahn I've tested those examples and agree that performance is impacted here by using vectorised rather than scalar. I have added checks to the LV patch in LLVM to not apply vectorisation to loops where the scalar cost is lower than the vectorisation cost. In your examples, these would now be excluded. 


I have added these here as examples of where it is not beneficial to vectorise this code as the pragma's still vectorise the loops.

Stylie777:
Ping

Stylie777:
Gentle Ping

MicroBenchmarks/LoopVectorization/SmallLoopTripCount.cpp
```diff
@@ -0,0 +1,259 @@
...
+  for (int i = 0; i < 3; i++) {
+    B[i] = B[i] < A[i] ? B[i] : A[i];
+  }
+}
+
+template <typename Ty>
+NOINLINE void loopTc5PowerVector(const Ty *__restrict A, Ty *__restrict B) {
+  LOOP_VECTORIZE_ENABLE
+  for (int i = 0; i < 5; i++) {
+    B[i] ^= A[i];
```
huntergr-arm, line 138: nit: in C(++), `^` is bitwise xor rather than exponentiation/power.

huntergr-arm (approved):
LGTM

MicroBenchmarks/LoopVectorization/SmallLoopTripCount.cpp
Stylie777: I've updated the function names here

PR #409, opened 2026-05-22
[AIX] Add compile-time max RSS reporting support on AIX.

sujianIBM:
This is a follow-up of PR #373. It adds compile-time max RSS reporting support on AIX.

### DIFF
--- a/tools/timeit.c
+++ b/tools/timeit.c
@@ -296,7 +296,7 @@ static int monitor_child_process(double start_time) {
   /* Apple reports max RSS in bytes */
   maxrss_bytes = (unsigned long) usage.ru_maxrss;
 #elif defined(__linux__) || defined(_AIX)
-  /* Linux reports max RSS in KiB */
+  /* Linux and AIX report max RSS in KiB */
   maxrss_bytes = (unsigned long) usage.ru_maxrss * 1024;
 #endif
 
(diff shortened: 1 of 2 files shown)


daltenty:
Shouldn't the configuration warning be modified as well if we are changing the supported platforms?
https://github.ibm.com/compiler/test-suite/blob/dd33df30587de389bcc6d4e0fd1971d813532602/CMakeLists.txt#L352

sujianIBM:
> Shouldn't the configuration warning be modified as well if we are changing the supported platforms? 

Sure! I will do this. Thanks!

sujianIBM:
Hello, I made the suggested change. Would you please take another look? Thanks!

tools/timeit.c
```diff
@@ -295,7 +295,7 @@ static int monitor_child_process(double start_time) {
 #if defined(__APPLE__)
   /* Apple reports max RSS in bytes */
   maxrss_bytes = (unsigned long) usage.ru_maxrss;
-#elif defined(__linux__)
+#elif defined(__linux__) || defined(_AIX)
   /* Linux reports max RSS in KiB */
```
daltenty: nit: update the comment to match
```suggestion
  /* Linux  and AIX report max RSS in KiB */
```

daltenty (approved):
LGTM, with minor nit

tools/timeit.c
sujianIBM: I made the suggested change. Thanks!

PR #410, opened 2026-05-23
[test-suite] Add musttail runtime regression tests for indirect args

xroche:
Twelve C programs under `SingleSource/Regression/C/musttail/` for `__attribute__((clang::musttail))` calls with struct-by-value arguments. The bug class: a tail-called function reads its arguments from caller-frame memory that the tail call has already deallocated. Static checks (assembly, IR shape) look right; only runtime execution diverges.

Argument values move through indirect pointers when the struct is too large for register passing. A normal call leaves the caller's frame alive; a musttail deallocates it before the callee runs. If the indirect pointer pointed into that frame, the callee reads freed memory.

The original seven tests target the bug class addressed by llvm/llvm-project#185094 (RISC-V backend): forwarded indirect, computed indirect, swapped args, cross-basic-block musttail, stack-spill, sret, mixed indirect + spill.

The five new tests target the broader class addressed by llvm/llvm-project#199351 (Clang frontend two-phase rewrite). Each uses a write-then-read-via-volatile probe to defeat the optimizer's noalias-by-ABI assumption that hides the bug at -O1:

- `alias-distinct-addresses` -- `C(a, a)`: per-slot storage must be distinct.
- `triple-alias` -- generalization to three slots.
- `local-source` -- `C(local)`: the source dangles unless routed through an incoming pointer.
- `mixed-sources` -- `C(local, a)`: both copy paths in one call.
- `swap-alias` -- `C(b, a, a)`: reorder combined with aliasing.

Cross-arch runtime validation: each new test fails with pre-#199351 clang on aarch64, riscv64, and s390x at -O2 and -O3 (the test-suite's typical optimization levels), and passes with the patched clang.

Gated on Clang (GCC has no `__attribute__((musttail))`). No changes outside the test suite.

Assisted-by: Claude (Anthropic)


xroche:
Pausing this PR.

On closer look only 2 of the 7 tests (sret-musttail, mixed-indirect-spill) actually distinguish pre- and post-merge clang. The other 5 pass on both: PR llvm/llvm-project#185094 fixes the backend's `CCValAssign::Indirect` path, which Clang's frontend bypasses for C struct-by-value by emitting an `alloca + memcpy + pass-pointer` (byval-temp).

That byval-temp is itself a real dangling-pointer bug on RV64 and AArch64 (covered by llvm/llvm-project#190429, #56435, #72555). The Clang fix is at llvm/llvm-project#199351; this PR is on hold until that lands, then I'll add the stack-clobber probe and revisit.

Assisted-by: Claude (Anthropic)

PR #411, opened 2026-05-26
[flang] Tests for rank-1 integer array bounds in explicit shape arrays

ivanrodriguez3753:


Fortran/cray/f2023/08/05/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+# Exclude tests that require unimplemented feature (conditional expression as
+# actual argument)
+file(GLOB _sources *.f90)
+
+set(Source ${_sources})
```
tarunprabhu: I am not sure what this is doing.

tarunprabhu (commented):
I have requested reviews from folks who are more familiar with F2023 features.

ivanrodriguez3753:
@akuhlens @eugeneepshteyn

kwyatt-ext:
I will review this soon.  It is in my queue.

Fortran/cray/f2023/08/05/declaration-explicit-array-bounds.f90
```diff
@@ -0,0 +1,274 @@
...
+!   4. Rank-1 lb with scalar ub (broadcast)
+!   5. SHAPE of assumed-shape (runtime, can't fold)
+!   6. 3D bounds
+!   7. Function result shaped by rank-1
+!   8. Loop re-evaluation (no stale state)
+!   9. Element write/read correctness
+!  10. Implied-do + array slices + arithmetic in bounds
+!  11. Vector subscripts in bounds expressions
+!  12. Block construct re-evaluation — bounds from enclosing scope
+
```
kwyatt-ext, line 18: I recommend adding a case with a UB < LB.  It should produce a valid array of 0 size, but could happen in dynamic code.

kwyatt-ext (changes requested):
It is a very well written test.  2 things to note:
1) Recommended case addition.
2) I didn't check but make sure code is merged before merging this.

Fortran/cray/f2023/08/05/declaration-explicit-array-bounds.f90
```diff
@@ -0,0 +1,274 @@
...
+!   3. Scalar lb with rank-1 ub (broadcast)
+!   4. Rank-1 lb with scalar ub (broadcast)
+!   5. SHAPE of assumed-shape (runtime, can't fold)
+!   6. 3D bounds
+!   7. Function result shaped by rank-1
+!   8. Loop re-evaluation (no stale state)
+!   9. Element write/read correctness
+!  10. Implied-do + array slices + arithmetic in bounds
+!  11. Vector subscripts in bounds expressions
+!  12. Block construct re-evaluation — bounds from enclosing scope
```
cenewcombe, line 12: A few more edge cases for your consideration:

- negative bounds
- modify/clobber the bounds variable within the same scope (verify that bounds do not change)
- declare a constant (PARAMETER) bounds array -- exercises the compile-time constant-folding path vs. the runtime extraction path

This comment is meant to be non-blocking. The test coverage is very good. It would also be reasonable to implement some of these checks as lit tests rather than in the test suite (or they may already be covered in lit).

cenewcombe (approved):
I left an inline comment for you to consider. I don't need to re-review, looks good!

kwyatt-ext (approved):
LGTM.  Nice and thorough test.

PR #412, opened 2026-05-29
[HIP] Make Python interpreter overridable

jplehr:
We have found that in certain circumstances (build and test hosts are different machines with different OS), one can run into a situation where llvm-lit encodes a specific Python version in its shebang. If that version is not available on the test host, it can suffice to run the llvm-lit script with a specific Python interpreter.

This patch enables us to do that for the External/HIP tests.

Put together AI-assisted.


mhalk (approved):
LGTM -- nice catch!

PR #414, opened 2026-06-05
Add missing <cstdlib> include to LoopVectorization/main.cpp

philnik777:


fhahn (commented):
Looks like this has already been fixed by the merged https://github.com/llvm/llvm-test-suite/pull/415

PR #415, opened 2026-06-05
[MicroBenchmarks] Add missing `<cstdlib>` include

kovdan01:
The `EXIT_SUCCESS` identifier is defined in `<cstdlib>`

See buildbot failure started occurring after libc++ has dropped transitive includes in https://github.com/llvm/llvm-project/pull/195509:

https://lab.llvm.org/buildbot/#/builders/229/builds/1990/steps/12/logs/stdio


vvereschaka (approved):
LGTM

fhahn (commented):
Thanks for the fix!

PR #416, opened 2026-06-05
[HIP] Add rocm-libraries and rocprim

jplehr:
This adds support to build and run tests for rocprim as the intial target library.

We want to be able to use llvm-test-suite to drive the build and test of some HIP libraries from the ROCm stack. Therefore, we parse TheRock submodule info about rocm-libraries version and pull-in the rocm- libraries version as currently tracked by TheRock submodule pointer.

The default for test parallelism is 16 processes which corresponds to the number of VMs allocated by the kernel driver per GPU.

### DIFF
--- a/External/HIP/CMakeLists.txt
+++ b/External/HIP/CMakeLists.txt
@@ -26,6 +26,66 @@ else()
 endif()
 include(${CMAKE_CURRENT_LIST_DIR}/HipCatchTests.cmake)
 
+# Get rocm-libraries commit from TheRock (for version alignment)
+function(get_therock_rocm_libraries_version OUT_VAR)
+  set(THEROCK_LOCAL_DEV_PATH "/tmp/TheRock" CACHE PATH "Local TheRock for dev")
+  set(THEROCK_REPO_URL "https://github.com/ROCm/TheRock.git" CACHE STRING "TheRock URL")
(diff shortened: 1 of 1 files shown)


External/HIP/CMakeLists.txt
```diff
@@ -26,6 +26,66 @@ else()
 endif()
 include(${CMAKE_CURRENT_LIST_DIR}/HipCatchTests.cmake)
 
+# Get rocm-libraries commit from TheRock (for version alignment)
+function(get_therock_rocm_libraries_version OUT_VAR)
```
jhuber6, line 30: Why do we need this, just to track the version? Couldn't that just be a variable the CI itself sets?

jplehr: This tracks the version of the libraries in TheRock, which is the version we care about. 
For now we want to track that version and reconsider when we use this in the upstream bots. This is for downstream right now. 
This version can change multiple times a day.

jhuber6: Is there not an easier way to manage this than doing it through CMake? Couldn't the CI just look it up and set it at configure time?

jplehr: I guess we could have a helper script that does it. But I would prefer that it is not separated across too many places so that it behaves similarly across machines somewhat easily.

jhuber6 (approved):

PR #419, opened 2026-06-09
[HIP] Test unsigned 23-bit division and remainder

LU-JOHN:
Test unsigned 23-bit division and remainder. The values chosen for the test were selected to hit failures in expandDivRem23Impl and demonstrates why
 https://github.com/llvm/llvm-project/pull/202753 is needed.


jplehr:
This still needs to be added to the CMake etc, right?

LU-JOHN:
> This still needs to be added to the CMake etc, right?

Added CMake and result file.

jplehr (approved):

PR #420, opened 2026-06-10
[HIP] Enable multi-version catch tests

jplehr:
Address various places where -${VARIANT_SUFFIX} appears to have been missing. This is a problem when running in an environment with multiple different ROCm versions installed.

### DIFF
--- a/External/HIP/CMakeLists.txt
+++ b/External/HIP/CMakeLists.txt
@@ -7,7 +7,16 @@ message(STATUS "TEST_SUITE_HIP_ROOT: ${TEST_SUITE_HIP_ROOT}")
 get_filename_component(HIP_CLANG_PATH ${CMAKE_CXX_COMPILER} DIRECTORY)
 message(STATUS "HIP_CLANG_PATH: ${HIP_CLANG_PATH}")
 
+# Options for enabling tests for external projects (Kokkos, Ginkgo, rocPRIM)
+option(EXTERNAL_HIP_TESTS_KOKKOS "Download, build, and test Kokkos HIP backend" OFF)
+option(EXTERNAL_HIP_TESTS_GINKGO "Download, build, and test Ginkgo HIP backend" OFF)
+option(EXTERNAL_HIP_TESTS_ROCPRIM "Download, build, and test rocPRIM" OFF)
+
+# Options to enable / disable specific test categories
+option(HIP_ENABLE_BLENDER_TEST "Enable Blender ray-tracing test (requires Blender 3.4+ and Python dependencies)" OFF)
+
 # Include Catch tests integration module
+include(${CMAKE_CURRENT_LIST_DIR}/HipCatchTests.cmake)
 
 # Find Python interpreter for running llvm-lit
 # The llvm-lit script may have a hardcoded Python version in its shebang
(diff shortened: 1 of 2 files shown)


jplehr:
There is one change that is not "just a fix": The patch introduces the `HIP_ENABLE_BLENDER_TEST` CMake variable. This needs to be set to even enable the whole HIP blender test logic that is inside CMakeList.txt.

External/HIP/CMakeLists.txt
```diff
@@ -198,16 +198,18 @@ macro(create_local_hip_tests VariantSuffix)
...
-  configure_file(workload/blender/test_blender.sh.in ${CMAKE_CURRENT_BINARY_DIR}/test_blender.sh @ONLY)
-  configure_file(workload/blender/verify_blender.sh.in ${CMAKE_CURRENT_BINARY_DIR}/verify_blender.sh @ONLY)
-  file(COPY utils/log_data.py DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
-  file(COPY utils/compare_image.py DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
-  file(COPY utils/requirements.txt DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
-  llvm_test_run(EXECUTABLE "/bin/bash" "test_blender.sh")
-  llvm_test_verify(/bin/bash verify_blender.sh %o)
-  llvm_add_test(blender.test test_blender.sh)
-  list(APPEND VARIANT_SIMPLE_TEST_TARGETS blender.test)
+  if (HIP_ENABLE_BLENDER_TEST)
```
jmmartinez, line 209: Is there going to be an `option(HIP_ENABLE_BLENDER_TEST  ...)`  ?

jmmartinez (commented):
I just left a minor comment. Otherwise, the changes look reasonable to me.

External/HIP/CMakeLists.txt
jplehr: Yes, there needs to be, I guess, but I got stuck testing a couple of other changes.

jmmartinez (approved):
Looks good to me.

External/HIP/CMakeLists.txt
```diff
@@ -7,7 +7,16 @@ message(STATUS "TEST_SUITE_HIP_ROOT: ${TEST_SUITE_HIP_ROOT}")
 get_filename_component(HIP_CLANG_PATH ${CMAKE_CXX_COMPILER} DIRECTORY)
 message(STATUS "HIP_CLANG_PATH: ${HIP_CLANG_PATH}")
 
+# Options for enabling tests for external projects (Kokkos, Ginkgo, rocPRIM)
+option(EXTERNAL_HIP_TESTS_KOKKOS "Download, build, and test Kokkos HIP backend" OFF)
+option(EXTERNAL_HIP_TESTS_GINKGO "Download, build, and test Ginkgo HIP backend" OFF)
+option(EXTERNAL_HIP_TESTS_ROCPRIM "Download, build, and test rocPRIM" OFF)
+
+# Options to enable / disable specific test categories
+option(HIP_ENABLE_BLENDER_TEST "Enable Blender ray-tracing test (requires Blender 3.4+ and Python dependencies)" OFF)
```
jplehr: Requiring this to `OFF` should be manually set, so it is NFC.
```suggestion
option(HIP_ENABLE_BLENDER_TEST "Enable Blender ray-tracing test (requires Blender 3.4+ and Python dependencies)" ON)
```

jplehr: I expect most environments having a single ROCm installation that is picked up, in which case this should be `ON`.

PR #423, opened 2026-06-12
[HIP] Test signed 23-bit division and remainder

LU-JOHN:
Test signed 23-bit division and remainder with a variety of values.  Specifically, 23-bit NEG_MAX divided by -1 is tested.  This shows why https://github.com/llvm/llvm-project/pull/203436 is needed.


jplehr:
Does the test work without the LLVM patch you are referencing or would it miscompile?

LU-JOHN:
> Does the test work without the LLVM patch you are referencing or would it miscompile?

It will miscompile without https://github.com/llvm/llvm-project/pull/203436.

jplehr (approved):
I think this can land now that the required LLVM patch has landed.

PR #425, opened 2026-06-15
[gfortran] Disable test incorrectly using assumed-size array

kparzysz:
This testcase declares an assumed-size dummy array and passes it to the DEPEND clause, expecting successful compilation. This is not allowed in OpenMP and fails to compile with flang.


kparzysz:
Similar errors with other compilers:
ifx: https://godbolt.org/z/cTWz6aTcs
ifort: https://godbolt.org/z/o7Yqcss8d

ceseo (approved):

PR #427, opened 2026-06-17
Revert "[flang] added write-after-endfile testcase"

kwyatt-ext:
Reverts llvm/llvm-test-suite#396

It looks like the expected results file might be formatted incorrectly.


kwyatt-ext:
@blazie2004 You will need to fix this as it is preventing the llvm-test-suite from running.  Please make sure you run your test in place in the whole suite before restoring your test to the suite.

PR #428, opened 2026-06-17
[HIP] Test 64-bit div with operands in 32-bit signed range

LU-JOHN:
Test 64-bit division with operands in a 32-bit signed range.


mgcarrasco:
LGTM, just in case, format the code if necessary.

mgcarrasco (approved):

PR #429, opened 2026-06-19
Switch ABI Test Suite to use the lit internal shell

dyung:
I tested using the internal shell on our bot and it seemed to work, so I'm going to enable it and see if our bot breaks.


sks75220:
Thanks

PR #430, opened 2026-06-23
Add write-after-endfile test under f2023/12/11

blazie2004:
This test verifies that Fortran I/O statements correctly handle error conditions when ERR= and IOSTAT= specifiers are present. Per Fortran standard 12.11.2, when an error occurs during I/O and either ERR= or IOSTAT= is specified.

PR: https://github.com/llvm/llvm-project/pull/191633


kwyatt-ext (approved):
LGTM

PR #431, opened 2026-06-29
[Polybench] Honour SMALL_PROBLEM_SIZE consistently

asb:
Some, but not all, of the Polybench benchmarks translate the SMALL_PROBLEM_SIZE CMake option to the SMALL_DATASET preprocessor define. This patch adds the necessary wiring for all those that didn't, and adds the corresponding small reference outputs. This makes a particular difference for lu, ludcmp, and heat-3d.


lukel97 (approved):
LGTM

PR #432, opened 2026-06-29
Add option to enable Debugify coverage-and-origin tracking builds

SLTozer:
To simplify the process of running debug location origin-tracking builds of the LLVM test suite, this patch adds a CMake flag and compiler wrapper script to enable the efficient collection of unexpectedly missing debug locations.

The wrapper script added by this patch first runs a build in coverage-only mode, which should run reasonably fast, and for each output object file where one or more missing debug locations were detected while building it, a coverage report is created at <object>.cov.json. If this file is created, the wrapper script reads it to see which passes created at least one missing location, and the re-runs the compile command with origin-tracking enabled for those specific passes, accumulating all results into <build>/debugify-report.json.

This flag is meant to be used with a Clang binary with coverage-and-origin tracking enabled; if only coverage tracking is enabled, the coverage report will be created at the same location without origin stacktraces (two builds will still be run even though one is redundant), and if there is no coverage or origin tracking then it will run a normal debugify build, still running redundant build commands in the process.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -347,6 +347,17 @@ if (TEST_SUITE_REPORT_COMPILE_MAX_RSS AND NOT (APPLE OR CMAKE_SYSTEM_NAME STREQU
   message(WARNING "TEST_SUITE_REPORT_COMPILE_MAX_RSS is only supported on Apple/Linux/AIX, ignored on this platform")
 endif()
 
+option(TEST_SUITE_COLLECT_DEBUGIFY_LOC_COVERAGE
+       "Check for missing debug location coverage (requires Clang with DebugLoc origin-tracking enabled)" ON)
+
+if (TEST_SUITE_COLLECT_DEBUGIFY_LOC_COVERAGE)
+  set(DEBUGIFY_ORIGIN_REPORT_PATH "${CMAKE_BINARY_DIR}/debugify-report.json")
+  set(DEBUGIFY_CMD_PREFIX "${CMAKE_SOURCE_DIR}/DebugifyCompilerWrapper.py ${DEBUGIFY_ORIGIN_REPORT_PATH}")
+
+  set(CMAKE_C_COMPILE_OBJECT "${DEBUGIFY_CMD_PREFIX} ${CMAKE_C_COMPILE_OBJECT}")
+  set(CMAKE_CXX_COMPILE_OBJECT "${DEBUGIFY_CMD_PREFIX} ${CMAKE_CXX_COMPILE_OBJECT}")
+endif()
+
 option(TEST_SUITE_COLLECT_COMPILE_TIME
        "Measure compile time by wrapping compiler invocations in timeit" ON)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -348,7 +348,7 @@ if (TEST_SUITE_REPORT_COMPILE_MAX_RSS AND NOT (APPLE OR CMAKE_SYSTEM_NAME STREQU
 endif()
 
 option(TEST_SUITE_COLLECT_DEBUGIFY_LOC_COVERAGE
-       "Check for missing debug location coverage (requires Clang with DebugLoc origin-tracking enabled)" ON)
+       "Check for missing debug location coverage (requires Clang with DebugLoc origin-tracking enabled)" OFF)
 
 if (TEST_SUITE_COLLECT_DEBUGIFY_LOC_COVERAGE)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -347,6 +347,17 @@ if (TEST_SUITE_REPORT_COMPILE_MAX_RSS AND NOT (APPLE OR CMAKE_SYSTEM_NAME STREQU
   message(WARNING "TEST_SUITE_REPORT_COMPILE_MAX_RSS is only supported on Apple/Linux/AIX, ignored on this platform")
 endif()
 
+option(TEST_SUITE_COLLECT_DEBUGIFY_LOC_COVERAGE
+       "Check for missing debug location coverage (requires Clang with DebugLoc origin-tracking enabled)" ON)
```
jmorse: Is this going to default to on for absolutely everyone using the llvm test suite? I imagine not everyone will desire this if so. (Possibly I'm missing some context).

DebugifyCompilerWrapper.py
```diff
@@ -0,0 +1,105 @@
+#!/bin/python3
+
+# CMake compiler wrapper which collects DebugLoc coverage tracking stats.
+# Runs a DebugLoc Coverage-Tracking compiler command first, checking to see
+# whether there is any missing coverage. If at least one missing location is
+# found, repeat the build with Origin-Tracking enabled and accumulate the
+# results into a single report file, shared across all compiler commands.
```
jmorse, line 7: Just to confirm an assumption, cmake isn't going to use this for absolutely /everything/ is it? It'd be infuriating to have a test-for-presence-of-header check fail and thus cmake fail overall, because a source location goes missing in the process.

DebugifyCompilerWrapper.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def get_source_and_output_files(args: List[str]):
+    source_file = None
+    output_file = None
+    for idx, arg in enumerate(args):
+        if arg == "-c" and idx + 1 < len(args):
+            source_file = args[idx + 1]
+        if arg == "-o" and idx + 1 < len(args):
+            output_file = args[idx + 1]
```
jmorse, line 24: Note that this is relying on cmakes convention for where it places the source file on the command line, we're not actually guaranteed that the argument after -c is the source file. (IMO it's fine to rely on stuff like this for our narrow scripting target).

jmorse (commented):
This broadly seems alright, although I'm not sure if I should approve as I've never submitted a patch to llvm-test-suite before. Do we have an estimate for how much faster this makes the coverage-tracking bot?

DebugifyCompilerWrapper.py
SLTozer: That's more or less my intention, i.e. rather than trying to parse out what is and isn't a source file argument in an arbitrary compile command, use the convention/ordering that is used for all C/C++ compile commands in llvm-test-suite.

CMakeLists.txt
SLTozer: Oops, yes - that should be OFF.

DebugifyCompilerWrapper.py
SLTozer: It will use it as the compiler wrapper for all C/C++ compile commands; it might apply to those checks, but if a source location goes missing that won't result in a failed compile. This wrapper only accumulates missing source locations into a report file, it never causes a passing compile command to fail - the failure on a buildbot comes from checking afterwards to see if any errors were submitted in the report. It would be rather odd if we got missing source location reports from a CMake header check compile, but they would at least be real errors, so I'm a bit on the fence about whether to try and circumvent that - the best/only way to do so would probably be to read the args of the compile command to determine whether it's a CMake-generated compile command.

jryans (commented):
Also looks broadly reasonable to me (assuming you change the option's default value as @jmorse mentioned).

SLTozer:
> Do we have an estimate for how much faster this makes the coverage-tracking bot?

Less than I originally thought*! It depends on whether or not there are any actual errors; in the case where there are no errors, it runs about 10% faster on average than a normal origin tracing build; in the case where there are lots of errors, it can go slower - around 33% with the current set of errors on the bot. Although 33>10, we expect that most of the time there will be no errors, so I think this would still be a net win overall. If we don't fancy the tradeoff though, we could compile just once with origin tracking - I'd still be in favour of including this flag so that we have a standard debugify build configuration in the test suite, which should help others do their own debug info testing.

*Turns out there were some substantial anti-synergies between origin tracking and the DILocations map that I'm attempting to abolish in https://github.com/llvm/llvm-project/pull/207374. The measurements I've put here are when that patch is turned on, which is what I expect the coverage bot will be using, since that is the largest win.

PR #434, opened 2026-06-30
[HashRecognize] Test naive/optimized against known reference output

xarkenz:
The current HashRecognize unit tests compare the results of hand-optimized and compiler-optimized CRC loops, and exit with either 0 or 1 to indicate pass or fail. To increase test reliability, compile each unit test twice (once unoptimized, once optimized for the current target) and have each print the results, then compare both outputs against a known good reference output.


SingleSource/UnitTests/HashRecognize/crc_macros.h
```diff
@@ -63,13 +27,14 @@
     }                                                                          \
   }
 
-// sample consists of random hand-picked values.
-#define VERIFY_RESULT(goldfn, actualfn, res)                                   \
+#define PRINT_RESULTS(fn, sample)                                              \
   {                                                                            \
-    static const unsigned sample[8] = {0, 1, 11, 16, 129, 142, 196, 255};      \
```
artagnon, line 69: Could keep the sample here, as opposed to duplicating it?

xarkenz: I had the idea that the sample inputs could be expanded to the specific bit widths being tested, but I can revert this change if that's not desirable, or perhaps go ahead and expand the samples.

artagnon: They will be implicitly widened anyway?

SingleSource/UnitTests/HashRecognize/CMakeLists.txt
```diff
@@ -1 +1,18 @@
+include(CheckCCompilerFlag)
+
 llvm_singlesource()
+
+function(llvm_hashrecognize_variant prefix)
```
pfusik: Will this function be used more than once?

SingleSource/UnitTests/HashRecognize/CMakeLists.txt
```diff
@@ -1 +1,18 @@
...
+  llvm_singlesource(PREFIX "${prefix}-" DEST_SUFFIX "-${prefix}")
```
pfusik: Please explain `DEST_SUFFIX "-${prefix}"`

pfusik (commented):
I like this change. Comparing against the code compiled without the transform makes the tests simpler and more robust.

SingleSource/UnitTests/HashRecognize/CMakeLists.txt
xarkenz: I had copied it from the Vectorizer cmake, but it seems to be unnecessary. Removing.

SingleSource/UnitTests/HashRecognize/CMakeLists.txt
xarkenz: I had based this off of Vectorizer, but since there isn't any foreseeable reason to use this more than once, I'll just replace the single usage and get rid of the function.

SingleSource/UnitTests/HashRecognize/crc_macros.h
xarkenz: Perhaps tests with larger CRC or data could test more bits, but that may not be in the scope of these changes. I'll go ahead and revert this change.

pfusik (approved):
LGTM

xarkenz:
@pfusik I don't have commit access, would you mind merging for me?

PR #435, opened 2026-07-01
[HIP] Add test with s_bitreplicate

vangthao95:
Add a test for s_bitreplicate when the inputs are non-uniform. The intrinsic is expanded to a VALU sequence.

Needs https://github.com/llvm/llvm-project/pull/208308.

### DIFF
--- a/External/HIP/CMakeLists.txt
+++ b/External/HIP/CMakeLists.txt
@@ -140,6 +140,9 @@ macro(create_local_hip_tests VariantSuffix)
   # Set per-source compilation/link options
   set_source_files_properties(with-fopenmp.hip PROPERTIES
     COMPILE_FLAGS -fopenmp)
+  # Use GlobalISel for device compilation
+  set_source_files_properties(s-bitreplicate.hip PROPERTIES
+    COMPILE_FLAGS "-fno-offload-lto -Xarch_device -fglobal-isel")
   # TODO: Add the flag after the kernel argument splitting pass is enabled.
   #set_source_files_properties(split-kernel-args.hip PROPERTIES
   #  COMPILE_FLAGS "-mllvm -amdgpu-enable-split-kernel-args")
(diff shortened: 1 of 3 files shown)


External/HIP/CMakeLists.txt
```diff
@@ -140,6 +140,9 @@ macro(create_local_hip_tests VariantSuffix)
   # Set per-source compilation/link options
   set_source_files_properties(with-fopenmp.hip PROPERTIES
     COMPILE_FLAGS -fopenmp)
+  # Use GlobalISel for device compilation
+  set_source_files_properties(s-bitreplicate.hip PROPERTIES
```
arsenm: We probably shouldn't be randomly enabling global isel for specific tests. It should either be a universal switch

External/HIP/s-bitreplicate.hip
```diff
@@ -0,0 +1,70 @@
...
+  const size_t bytes = N * sizeof(uint64_t);
+
+  uint64_t *d_out;
+  HIP_CHECK(hipMalloc((void **)&d_out, bytes));
+
+  const int threads = 256;
+  bitreplicate_kernel<<<(N + threads - 1) / threads, threads>>>(d_out, N);
+  HIP_CHECK(hipDeviceSynchronize());
+
+  uint64_t *h_out = (uint64_t *)malloc(bytes);
```
arsenm: C++ smart pointers

External/HIP/CMakeLists.txt
jplehr: Yeah, I think it would be great if we can enable the whole test suite to run with GlobalISel.

vangthao95: Agreed it would be ideal

vangthao95:
Changed to using SDAG equivalent: https://github.com/llvm/llvm-project/pull/208308

jplehr (approved):
I think this looks reasonable. 

Then we can introduce a cmake switch to use GlobalISel across all tests.

PR #436, opened 2026-07-01
[Vectorizer] Extend fpmin/fpmax tests to test semantic variants

nasherm:
This extends the test cases in argmin-argmax-fp to test for the semantic variants of fpmin/fpmax operations. This test will now also cover fpminimum/fpmaximum and fpminimum_num/fpmaximum_num.


SingleSource/UnitTests/Vectorizer/argmin-argmax-fp.cpp
```diff
@@ -90,7 +145,55 @@
   T_FMAX_ADD("max", M, I, 0, 1, std::numeric_limits<M>::lowest(), 0,           \
              "_fmax_add_start_0")                                              \
   T_FMAX_ADD("max", M, I, 0, 2, std::numeric_limits<M>::lowest(), 0,           \
-             "_fmax_add_start_0_inc_2")
+             "_fmax_add_start_0_inc_2")                                        \
+  T_FMINIMUM("minimum", M, I, 0, 1, std::numeric_limits<M>::max(), 0,          \
```
fhahn, line 149: I think the minimum/maxium are missing cases with `NaN`? I think `NaN` inputs may currently be miscompiled with the PR as per https://github.com/llvm/llvm-project/pull/198300#discussion_r3684465737. Would be good to make sure we have interesting inputs with NaN covered for maximum/minimum

nasherm: Done

nasherm:
Ping. Completely forgot about this patch

PR #437, opened 2026-07-03
[compare.py] Add --only-changed option

lukel97:
I plan on using the ./compare.py script for the /test-suite workflow in llvm-project, and as a part of that it would be useful to only show results that have actually changed for a given metric. So this adds a flag for that, --only-changed. It should work across multiple sets of results, including 3 or more.

I used Claude to help reason about the Pandas stuff.

### DIFF
--- a/utils/compare.py
+++ b/utils/compare.py
@@ -374,6 +374,7 @@ def print_result(
     shorten_names=True,
     minimal_names=False,
     show_diff_column=True,
+    only_changed=False,
     sortkey="diff",
     sort_by_abs=True,
     absolute_diff=False,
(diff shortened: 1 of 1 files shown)


wangpc-pp (approved):
LGTM. Sounds useful!

tomershafir:
I wonder if `--statistics --only-significant` could have also work

lukel97:
> I wonder if `--statistics --only-significant` could have also work

I had this thought as well, but from what I understand that's based off the p-value and I couldn't get it to work with `--alpha 0.0`. Since I want to specifically show any deviation in a metric, not just those past a certain threshold (also without necessarily having to print the statistics column)

PR #438, opened 2026-07-06
[compare.py] Add flag for outputting in markdown and csv format

lukel97:
In order to display the results of compare.py in a github comment from the /test-suite workflow, this patch adds a --format=md flag.

It's pretty easy to support other formats too so I've added in a csv formatter too, it's useful for creating graphs and charts etc.

Some of the other print calls were moved to stderr so it doesn't clobber the formatted output on stdout.

Here's an example table produced below from `/utils/compare.py foo.json vs bar.json -m size..text --only-changed  --minimal-names --lhs-name base --rhs-name head --format=md`:

| ('Program', '')    |   ('size..text', 'base') |   ('size..text', 'head') |   ('size..text', 'diff') |
|:-------------------|-------------------------:|-------------------------:|-------------------------:|
| queens             |                     1144 |                     1170 |              0.0227273   |
| fldry              |                      882 |                      902 |              0.0226757   |
| dry                |                      774 |                      788 |              0.0180879   |
| mst                |                     1850 |                     1876 |              0.0140541   |
| IRSmk              |                     5236 |                     5302 |              0.012605    |
| spiff              |                    15402 |                    15572 |              0.0110375   |
| power              |                     4544 |                     4592 |              0.0105634   |
| Shootout-sieve     |                      404 |                      408 |              0.00990099  |
| himenobmtxpa       |                     6356 |                     6412 |              0.00881057  |
| ray                |                     3226 |                     3254 |              0.00867948  |
| jacobi-2d          |                     2806 |                     2824 |              0.00641483  |
| richards_benchmark |                     2718 |                     2734 |              0.00588668  |
| sim                |                    11232 |                    11296 |              0.00569801  |
| syr2k              |                     2476 |                     2490 |              0.00565428  |
| HPCCG              |                    14598 |                    14676 |              0.0053432   |
| Geomean difference |                      nan |                      nan |              0.000118887 |

There's probably nicer ways to format the floats/geomean difference but I figured it would be good to just get the basics plugged in for now.

### DIFF
--- a/utils/compare.py
+++ b/utils/compare.py
@@ -26,8 +26,7 @@ def read_lit_json(filename):
     info_columns = ["hash"]
     # Pass1: Figure out metrics (= the column index)
     if "tests" not in jsondata:
-        print("%s: Could not find toplevel 'tests' key" % filename,
-              file=sys.stderr)
+        print("%s: Could not find toplevel 'tests' key" % filename, file=sys.stderr)
         sys.exit(1)
     for test in jsondata["tests"]:
         name = test.get("name")
@@ -38,8 +37,10 @@ def read_lit_json(filename):
             sys.stderr.write("Error: Multiple tests with name '%s'\n" % name)
             sys.exit(1)
         if "metrics" not in test:
-            print("Warning: '%s' has no metrics, skipping!" % test["name"],
-                  file=sys.stderr)
+            print(
+                "Warning: '%s' has no metrics, skipping!" % test["name"],
+                file=sys.stderr,
+            )
             continue
         names.add(name)
         for name in test["metrics"].keys():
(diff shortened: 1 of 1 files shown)


utils/compare.py
```diff
@@ -647,6 +656,11 @@ def main():
         default=False,
         help="Show only results where a metric has changed"
     )
+    parser.add_argument(
+        "--format",
+        choices=['csv', 'md'],
```
wangpc-pp: Add a new choice `raw` or `str` and make it default to represent the default behavior?

lukel97: Good point, added `text` in d15fabd74f432f3df57a874e9855dacb3beab3d0

wangpc-pp (approved):
LGTM.

PR #439, opened 2026-07-08
[Fortran/gfortran] Disable bounds_check_fail_3.f90 test

kaviya2510:
The test `bounds_check_fail_3.f90` started failing at -O1/-O2/-O3 after https://github.com/llvm/llvm-project/pull/208159
This fix extends `hlfir.assign` inlining to handle assignments where both the LHS and RHS are slices of the same array, as well as array assignments that may alias with other arrays, so the runtime is not called and it doesn't make runtime bounds checks.


tarunprabhu (approved):
Could you keep that list of files sorted please. So add the entry for `bounds_check_fail_3.f90` immediately after `all_bounds_1.f90`. It just makes it a bit easier to find things in an already somewhat disorganized file. Thanks.

jeanPerier (approved):

tblah (approved):

PR #440, opened 2026-07-08
Fix telecomm-FFT and beamformer benchmarks on GCC

lukel97:
These benchmarks use #pragma STDC FP_CONTRACT OFF, but this isn't supported in GCC: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=20785#c7

Instead use a gcc specific function pragma when the compiler is GCC. We still want to allow FP contraction in the main part of the test.

Claude was used to help diagnose the failure.

### DIFF
--- a/MultiSource/Benchmarks/MiBench/telecomm-FFT/fourierf.c
+++ b/MultiSource/Benchmarks/MiBench/telecomm-FFT/fourierf.c
@@ -168,6 +168,13 @@ void fft_float (
     }
 }
 
+// GCC doesn't support the FP_CONTRACT pragma.
+#if defined(__GNUC__) && !defined(__clang__)
+#pragma GCC optimize ("fp-contract=off")
+#else
+#pragma STDC FP_CONTRACT OFF
+#endif
+
 void fft_float_StrictFP (
     unsigned  NumSamples,
     int       InverseTransform,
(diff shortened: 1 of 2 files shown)


asb (approved):
Minor comments:
I think "apply the pragma at the function level in beamformer" is a bit confusing as worded as "function level" to me could also apply the previous arrangement. Perhaps "apply the pragma across a region of multiple functions"?

There's no correctness issue here, but I wondered if it would be better 'hygiene' to explicitly end the region of fp-contract-off with pop_options, but then I'm unclear if `#pragma STDC FP_CONTRACT DEFAULT` actually has the same semantics if ffp-contract is set to something explicitly on the command line.

lukel97:
> Minor comments: I think "apply the pragma at the function level in beamformer" is a bit confusing as worded as "function level" to me could also apply the previous arrangement. Perhaps "apply the pragma across a region of multiple functions"?
> 
> There's no correctness issue here, but I wondered if it would be better 'hygiene' to explicitly end the region of fp-contract-off with pop_options, but then I'm unclear if `#pragma STDC FP_CONTRACT DEFAULT` actually has the same semantics if ffp-contract is set to something explicitly on the command line.

I've switched this to use the function pragma you mentioned in https://github.com/llvm/llvm-test-suite/pull/441#issuecomment-4969137962 instead, so we can avoid clobbering the region

PR #441, opened 2026-07-08
Fix Polybench benchmarks using FP_CONTRACT pragma on GCC

lukel97:
Similar to #440, this implements the same fix for any polybench test that relies on #pragma STDC FP_CONTRACT OFF, which isn't supported on GCC and causes mismatched results.

Outside of these two PRs there doesn't seem to be any other users of the pragma.

The changes in this PR were written by Claude

### DIFF
--- a/SingleSource/Benchmarks/Polybench/datamining/correlation/correlation.c
+++ b/SingleSource/Benchmarks/Polybench/datamining/correlation/correlation.c
@@ -22,10 +22,7 @@
 
 
 /* Array initialization. */
-#if defined(__GNUC__) && !defined(__clang__)
-#pragma GCC push_options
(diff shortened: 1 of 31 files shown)


asb:
It looks like GCC supports function attribute ` __attribute__((optimize("fp-contract=off")))`. Did you evaluate doing that rather than push+pop?

lukel97:
> It looks like GCC supports function attribute ` __attribute__((optimize("fp-contract=off")))`. Did you evaluate doing that rather than push+pop?

That looks cleaner, done in f05d957c6bd22db34061d7479fa25812c670be6e thanks

asb (approved):
LGTM

PR #446, opened 2026-07-13
Define UNALIGNED_MEM_ACCESS_NOT_SUPPORTED in ALAC test for RISC-V when needed

kito-cheng:
ag_enc.c casts uint8_t* to uint32_t* to read and write big-endian 32-bit values. The compiler assumes a uint32_t* is aligned and emits a full-word load/store, which traps on RISC-V without misaligned load/store support.

Enable it for RISC-V targets whose compiler defines __riscv_misaligned_avoid, so targets that support misaligned accesses keep using the word-sized path.


topperc (approved):
LGTM

wangpc-pp (approved):

lukel97 (approved):

PR #447, opened 2026-07-14
[Fortran] declare-variant-6 is now passing after LLVM's PR #206988

pawosm-arm:
See this PR: https://github.com/llvm/llvm-project/pull/206988

### DIFF
--- a/Fortran/gfortran/regression/gomp/tests.cmake
+++ b/Fortran/gfortran/regression/gomp/tests.cmake
@@ -171,7 +171,7 @@ compile;declare-variant-2a.f90;xfail;;;
 compile;declare-variant-3.f90;;;;
 compile;declare-variant-4.f90;;;;
 compile;declare-variant-5.f90;;;i.86-.+-.+ x86_64-.+-.+;
-compile;declare-variant-6.f90;xfail;;;
+compile;declare-variant-6.f90;;;;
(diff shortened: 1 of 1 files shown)


abidh:
Previously we emitted a "not yet implemented" error for this directive, so the compile produced an error. After declare-variant support landed, that error is gone and the file now compiles cleanly, so switching the test to pass is fine for now.

It is worth noting though that the actual restriction this test checks is not yet implemented in our front end. The relevant rule is OpenMP 6.0, Section 9.6 (Declare Variant Directives, Restrictions):

"If a procedure is determined to be a function variant through more than one declare variant directive then the construct selector set of their context selectors must be the same."

I am working on it and when it lands, we can turn it back it back to xfail.

abidh (approved):

pawosm-arm:
considering we have failing CI on the day of branching, I need to merge it.

tarunprabhu:
Please revert this. If a test is to be temporarily disabled, it should be added to an appropriate list in `DisabledFiles.cmake`. There is a banner at the top of `tests.cmake` that says that the file should not be modified.

The `tests.cmake` files are generated by a script and will be overwritten - especially when these tests are sync'ed with upstream.

pawosm-arm:
> Please revert this.

See PR #448

PR #450, opened 2026-07-20
Use unique temp file for stripped executable in hash.py to avoid races

msolk:
Concurrent test workers can share the same executable path; using a per-invocation temp file (instead of a fixed `.stripped` suffix) prevents intermittent "No such file or directory" errors when one worker strips/cleans up while another reads the same path.

### DIFF
--- a/litsupport/modules/hash.py
+++ b/litsupport/modules/hash.py
@@ -4,6 +4,7 @@ import hashlib
 import logging
 import os
 import platform
+import tempfile
 
 
 def compute(context):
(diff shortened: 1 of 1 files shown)


msolk:
Request review from @MatzeB

msolk:
@MatzeB Gentle Reminder. Please let me know if I should request someone else for this.

msolk:
Request review from @wangpc-pp

wangpc-pp:
The code looks good to me. I am just wondering when one worker will strip/clean up the executable. I haven't met such error before.

msolk:
> The code looks good to me. I just am wondering when one worker will strip/clean up the executable. I haven't met such error before.

May be it only happens with certain config, we are seeing it in our downstream CI builds. Shall I try to create a small example trying to reproduce the error?

wangpc-pp:
> > The code looks good to me. I just am wondering when one worker will strip/clean up the executable. I haven't met such error before.
> 
> May be it only happens with certain config, we are seeing it in our downstream CI builds. Shall I try to create a small example trying to reproduce the error?

Yes. If the upstream can't reproduce the error, then we'd better to keep this change downstream. :-)

lukel97:
I haven't seen this before on our LNT configuration but I guess this makes sense, on LNT we run the benchmarks with -j1 so we wouldn't get the race conditions 

Do you know what benchmarks/tests we have that share an executable within the test-suite?

msolk:
> I haven't seen this before on our LNT configuration but I guess this makes sense, on LNT we run the benchmarks with -j1 so we wouldn't get the race conditions
> 
> Do you know what benchmarks/tests we have that share an executable within the test-suite?

it seems all the errors are for a dummy executable example `Fortran/gfortran/regression/dummy.stripped`

there are about 421 such errors

lukel97:
> > I haven't seen this before on our LNT configuration but I guess this makes sense, on LNT we run the benchmarks with -j1 so we wouldn't get the race conditions
> > Do you know what benchmarks/tests we have that share an executable within the test-suite?
> 
> it seems all the errors are for a dummy executable example `Fortran/gfortran/regression/dummy.stripped`
> 
> there are about 421 such errors

Maybe a more reliable way to fix this is to avoid the dummy executable altogether? it seems like it's only created to avoid  some sort of limitation of `llvm_test_run`, see `gfortran_add_compile_test`

msolk:
> > > I haven't seen this before on our LNT configuration but I guess this makes sense, on LNT we run the benchmarks with -j1 so we wouldn't get the race conditions
> > > Do you know what benchmarks/tests we have that share an executable within the test-suite?
> > 
> > 
> > it seems all the errors are for a dummy executable example `Fortran/gfortran/regression/dummy.stripped`
> > there are about 421 such errors
> 
> Maybe a more reliable way to fix this is to avoid the dummy executable altogether? it seems like it's only created to avoid some sort of limitation of `llvm_test_run`, see `gfortran_add_compile_test`

I can try to remove the limitation on `llvm_test_run`. However, from a cursory analysis. it means we will have to do some fundamental changes like removing "missing executables check" from `litsupport/test.py` and "RUN: line check" from `litsupport/testfile.py`. Would that be ok?

PR #454, opened 2026-07-21
[HIP] Support testing with multiple sets of options

LU-JOHN:
Support HIP testing with multiple sets of options specified in a <testcase>.options file.

Reference output is found in <TEST_NAME>-<VARIANT_NUMBER>.reference_output if it exists.  Otherwise reference output is found in <TEST_NAME>.reference_output.  For example, the first option set for a test named foo.hip would check foo-0.reference_output if it exists.  Otherwise foo.reference_output will be checked.

Used to support testing fast-i24-div-path.hip with -O0 and -O3 options.

-O0 is necessary to test LowerDIVREMToFloat in AMDGPUISelLowering.cpp
-O3 is necessary to test expandDivRemToFloatImpl in AMDGPUCodeGenPrepare.cpp

Testcase multi_output.hip illustrates testing with reference files that depend on the compile options.  In this testcase, option sets [0,3] use the default multi_output.reference_output.  Option sets [4,8] use the option set specific multi_output-\<N\>.reference_output.


yxsamliu:
I think this should keep checking the existing reference output file. With the new names, the tests will first look for fast-i24-div-path-0.reference_output / fast-i24-div-path-1.reference_output, so fast-i24-div-path.reference_output is no longer used.

Could we first try the postfixed reference output name, then fall back to the original non-postfixed name? That would allow per-option expected output while keeping the common-output case working.

LU-JOHN:
> I think this should keep checking the existing reference output file. With the new names, the tests will first look for fast-i24-div-path-0.reference_output / fast-i24-div-path-1.reference_output, so fast-i24-div-path.reference_output is no longer used.
> 
> Could we first try the postfixed reference output name, then fall back to the original non-postfixed name? That would allow per-option expected output while keeping the common-output case working.

@yxsamliu, I've added support for checking an option set specific reference file if it exists, with a fallback to the default reference file.

yxsamliu (approved):
LGTM. The fallback to the common reference output handles the main compatibility concern.

PR #455, opened 2026-07-27
[acc] Update OpenACC failing test categorization

razvanlupusoru:
After https://github.com/llvm/llvm-project/pull/211894 flang now emits not-yet-implemented messages for the unsupported constructs. This PR updates the categories for the tests.

It also ensures cache-1 test is disabled which caused the suite to fail.


razvanlupusoru:
I also create a small change to allow me to merge immediately: https://github.com/llvm/llvm-test-suite/pull/456
And then this PR can be properly reviewed.

tarunprabhu (approved):
Thanks for updating this. LGTM.

PR #457, opened 2026-07-31
[Fortran] Add namelist/ UnitTest for the empty-scalar extension

ergawy:
Covers Flang's NAMELIST extension that accepts an empty assignment to a scalar item (e.g. `l_flag=` immediately followed by the next name-value pair or the group terminator), leaving the item at its current value.  See flang/docs/Extensions.md.

The single test file exercises both spellings — `l_flag= i_count=7` with no whitespace around `=`, and `l_flag = i_count = 7` with whitespace on both sides — and uses `stop N` to signal per-check failure, following the pattern of the sibling UnitTests (binary_edit, fcvs21_f95, ...).  The directory is named `namelist/` (not `namelist_empty_scalar/`) so future NAMELIST extension tests can accrete alongside without another CMake subdirectory.

### DIFF
--- a/Fortran/UnitTests/namelist/namelist_empty_scalar.f90
+++ b/Fortran/UnitTests/namelist/namelist_empty_scalar.f90
@@ -1,0 +1,36 @@
+! Flang NAMELIST extension: an assignment to a scalar item may omit its
+! value (e.g. `l_flag=`, immediately followed by the next name-value
+! pair or the group terminator), leaving the item's current value
+! unchanged.  F2023 13.11.3.2 requires a value token to follow the `=`,
+! but nvfortran and gfortran accept the empty form as "keep current
+! value".  See flang/docs/Extensions.md.
+!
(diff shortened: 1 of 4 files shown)


mleair (commented):
LGTM

tarunprabhu (approved):
LGTM. Thanks.

Fortran/UnitTests/namelist/namelist_empty_scalar.f90
```diff
@@ -0,0 +1,36 @@
+! Flang NAMELIST extension: an assignment to a scalar item may omit its
+! value (e.g. `l_flag=`, immediately followed by the next name-value
+! pair or the group terminator), leaving the item's current value
+! unchanged.  F2023 13.11.3.2 requires a value token to follow the `=`,
```
eugeneepshteyn: 13.11.2 p1, like in the original PR.

PR #458, opened 2026-07-31
[HashRecognize] Avoid CMake race condition when copying reference output

xarkenz:
The two tested variants of the HashRecognize tests (HashRecognize enabled/disabled) use different test prefixes, but use the same reference output file. This seems to be causing spurious failures with LLVM BuildBot in a manner very similar to what was seen in #381. To fix, simply add a `DEST_SUFFIX` as introduced in the aforementioned PR, so that each test copies its reference output to a distinct path.

Assisted-by: Claude


pfusik (approved):
LGTM

PR #459, opened 2026-08-03
[HashRecognize] Add variants for each CRC strategy

xarkenz:
Now that LLVM allows specifying a strategy for optimizing CRC loops with `-loop-idiom-crc-strategy`, run two more variants of HashRecognize for `table` and `clmul`. To simplify this process, the existing CMake logic has been converted to a function.


artagnon (commented):
I'm not sure this is desired, but I'll let Piotr take a look?

xarkenz:
Gentle ping.

SingleSource/UnitTests/HashRecognize/CMakeLists.txt
```diff
@@ -2,13 +2,18 @@ include(CheckCCompilerFlag)
...
-
-    list(APPEND CFLAGS "-mllvm" "-disable-loop-idiom-hashrecognize")
-    llvm_singlesource(PREFIX "no-hr-" DEST_SUFFIX "-no-hr")
+    list(APPEND CFLAGS "-mllvm" ${llvm_flag})
+    llvm_singlesource(PREFIX "${variant_name}-" DEST_SUFFIX "-${variant_name}")
     set(CFLAGS ${BASE_CFLAGS})
   endif()
+endfunction()
+
+if ("${CMAKE_C_COMPILER_ID}" MATCHES "Clang")
```
artagnon: We're checking check_c_compiler_flag in llvm_hashrecognize_variant? Would it be possible to get rid of this?

xarkenz: True, I guess `check_c_compiler_flag` already prevents this from being run on non-LLVM compilers. Didn't consider that, I had mostly been going off of some of the other CMake configurations nearby. I think it's safe to remove it here.

artagnon (approved):
LGTM, thanks.

PR #460, opened 2026-08-05
[HIP] Avoid symlink collision

LU-JOHN:
Multiple test option variants can share the same base reference output file.
Include the test option number in the suffix to avoid race condition where
parallel builds try to create the same symlink (e.g., test-0, test-1, test-2
all trying to create test.reference_output-hip-7.1.1).


LU-JOHN:
Issue introduced in https://github.com/llvm/llvm-test-suite/pull/454.

jplehr (approved):

jplehr:
Maybe change the title to include `[HIP]` to make clear for what this change is.

PR #461, opened 2026-08-07
[SPEC] Fix gobmk with -fwrapv-pointer

zqb-all:
Fixes pointer arithmetic UB that causes failures with recent clang.

Similar to xalancbmk fix in 0c27673ac.

https://discourse.llvm.org/t/is-anyone-else-seeing-an-error-in-spec2k6-445-gobmk-using-just-scalar/85752


lukel97 (approved):
LGTM

PR #462, opened 2026-08-11
[Fortran] Add a test for BIND(C) derived types passed by value

ergawy:
Check that a BIND(C) derived type dummy argument with the VALUE attribute is passed correctly to a C function, for sizes that are small enough to be passed in registers and for sizes that are not. Also check that arguments following such a derived type are passed correctly, since a derived type that is passed as a pointer to a copy still uses an argument register.

This is a run test that calls into C, so it covers both the caller and the callee side of the interface.

The AArch64 case is fixed by https://github.com/llvm/llvm-project/pull/215508.

### DIFF
--- a/Fortran/UnitTests/bind-c-value/CMakeLists.txt
+++ b/Fortran/UnitTests/bind-c-value/CMakeLists.txt
@@ -1,3 +1,7 @@
-llvm_multisource(bind-c-value)
+# BIND(C) is not implemented for x86_64 Windows, so do not enable the test on
+# Windows yet.
+if (NOT WIN32)
+  llvm_multisource(bind-c-value)
 
-file(COPY lit.local.cfg DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
+  file(COPY lit.local.cfg DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
+endif()
--- a/Fortran/UnitTests/bind-c-value/CMakeLists.txt
+++ b/Fortran/UnitTests/bind-c-value/CMakeLists.txt
@@ -1,0 +1,3 @@
+llvm_multisource(bind-c-value)
+
+file(COPY lit.local.cfg DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
--- a/Fortran/UnitTests/CMakeLists.txt
+++ b/Fortran/UnitTests/CMakeLists.txt
@@ -8,3 +8,4 @@ add_subdirectory(fcvs21_f95) # NIST Fortran Compiler Validation Suite
 add_subdirectory(finalization)
 add_subdirectory(fp_conversions)
 add_subdirectory(namelist)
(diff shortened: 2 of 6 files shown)


DavidTruby (commented):
As I recall `BIND(C)` is not fully implemented on x86_64-pc-windows-msvc so I think this test will fail there, although I don't have a machine to test it on at the moment (I'll try to test it today or tomorrow). It's possible we'll just need to XFAIL or otherwise exclude this on that triple.

tarunprabhu (approved):
Looks reasonable to me. I've asked @kwyatt-ext to take a look as well in case there are some subtleties of Fortran that are also worth checking for.

ergawy:
Thanks for the review @DavidTruby and @tarunprabhu. I will hold off merging for today in case you have a chance to further test it.

I don't have access to a Windows machine atm.

kwyatt-ext:
I can review this, but I will not be able to get to this until next week.

DavidTruby:
Ok I got onto a Windows on x86_64 machine and managed to test this; it does fail to build there because nobody has implemented `BIND(C)` on x86_64 Windows. So this will need to be XFAIL'd on `x86_64-pc-windows-msvc` and `x86_64-w64-windows-gnu`.

ergawy:
> Ok I got onto a Windows on x86_64 machine and managed to test this; it does fail to build there because nobody has implemented `BIND(C)` on x86_64 Windows. So this will need to be XFAIL'd on `x86_64-pc-windows-msvc` and `x86_64-w64-windows-gnu`.

Thanks David.

I disabled the test Windows the same as other tests that do so:
 - Fortran/cray/OpenMP/CMakeLists.txt
 - gfortran tests in Fortran/CMakeLists.txt

tarunprabhu:
Thanks Kareem. It might be worth moving the `if (WIN32)` check up a level and not entering the directory at all in that case. It might make it easier to find tests to re-enable when support for Windows is added. What do you think?

DavidTruby:
This test does pass on aarch64-pc-windows-msvc and arm64ec-pc-windows-msvc so I'd rather it wasn't disabled on Windows unconditionally. Would it be possible to disable only for X86 Windows?

Fortran/UnitTests/bind-c-value/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+# BIND(C) is not implemented for x86_64 Windows, so do not enable the test on
+# Windows yet.
+if (NOT WIN32)
```
DavidTruby, line 3: I think you can limit to x86_64 with `if (NOT (WIN32 AND CMAKE_SYSTEM_PROCESSOR STREQUAL "AMD64"))` or similar

kwyatt-ext (commented):
I agree with @DavidTruby 's comment.  Narrow the scope of exclusion.  Otherwise LGTM.

PR #463, opened 2026-08-12
[HashRecognize] Add carry-based CRC-16 test

vkortbeek-gf:
Add a little-endian CRC-16 test that computes the polynomial and carry updates separately. The bit test in the code is canonicalized to `trunc X to i1`, exercising HashRecognize's handling of that pattern.

This covers the source pattern from llvm/llvm-project#213881 and accompanies llvm/llvm-project#213883.


vkortbeek-gf:
@artagnon  I've implemented a basic test, but I'm not sure if this is what you expected. I did not use the macros like the other tests because getting the `trunc ... to i1` from the regression requires a different spelling of the CRC function. I can add a special `_CARRY` variant, but I'm not sure that would be useful for a one-off test. I also saw some precedent for not using the macros in the `.nodata` tests.

Let me know what I can improve :)

SingleSource/UnitTests/HashRecognize/crc16.le.data8.carry.c
xarkenz, line 1: After some testing, it seems to me that this code does produce `trunc X to i1`, but due to the extra `or` at the end, HashRecognize rejects it, and so the test becomes trivial. It seems pretty difficult to reproduce the trunc in a way that HashRecognize accepts, which is probably why this has flown under the radar.

One thing that seems promising is using the clang `_BitInt` extension to force the `i1` type. Maybe something like that can be used? This test would need to only run under clang, then.

xarkenz: Oops, I didn't notice https://github.com/llvm/llvm-project/pull/213882 before I made this comment. Feel free to disregard.

SingleSource/UnitTests/HashRecognize/crc16.le.data8.carry.c
```diff
@@ -0,0 +1,25 @@
...
+#include <stdint.h>
+
+#define GENPOLY 0x4002
+
+static NOINLINE uint16_t crc_loop(uint16_t crc_initval, uint8_t data) {
+  uint16_t crc = crc_initval;
+
+  // This carry-based loop will be optimized by HashRecognize.
+  for (uint8_t i = 0; i < 8; ++i) {
+    uint8_t x16 = (uint8_t)((data & 1) ^ ((uint8_t)crc & 1));
```
artagnon:
```suggestion
    uint8_t x16 = (data & 1) ^ (crc & 1);
```

I don't think the casts add much value? C would implicitly convert?

SingleSource/UnitTests/HashRecognize/crc16.le.data8.carry.c
```diff
@@ -0,0 +1,25 @@
+#include "crc_macros.h"
+#include <stdint.h>
+
+#define GENPOLY 0x4002
+
+static NOINLINE uint16_t crc_loop(uint16_t crc_initval, uint8_t data) {
+  uint16_t crc = crc_initval;
+
+  // This carry-based loop will be optimized by HashRecognize.
```
artagnon: Could add a brief comment about the alternate spelling due to InstCombine?

artagnon (approved):
LGTM, thanks.

SingleSource/UnitTests/HashRecognize/crc16.le.data8.carry.c
```diff
@@ -0,0 +1,25 @@
...
+  for (uint8_t i = 0; i < 8; ++i) {
+    uint8_t x16 = (uint8_t)((data & 1) ^ ((uint8_t)crc & 1));
+    data >>= 1;
+    uint8_t carry = (x16 == 1) ? 1 : 0;
```
artagnon:
```suggestion
    uint8_t carry = x16 == 1;
```

SingleSource/UnitTests/HashRecognize/crc16.le.data8.carry.c
```diff
@@ -0,0 +1,25 @@
...
+    crc = (x16 == 1) ? (crc ^ GENPOLY) : crc;
```
artagnon, line 16:
```suggestion
    crc = carry ? (crc ^ GENPOLY) : crc;
```

vkortbeek-gf: This one is needed for the `trunc to i1` to appear. I did move it added a comment and added an explanation comment above the loop. So it should be more clear (because it indeed looks quite strange).

SingleSource/UnitTests/HashRecognize/crc16.le.data8.carry.c
vkortbeek-gf: I've added a comment that should explain the situation.

xarkenz (approved):
LGTM, thanks. Since the test is logically correct, merging this doesn't need to depend on https://github.com/llvm/llvm-project/pull/213882 and https://github.com/llvm/llvm-project/pull/213883, but FWIW, when I temporarily apply both of those changes and compile the new test, it gets properly detected and optimized.

[print_before_after.txt](https://github.com/user-attachments/files/31035348/print_before_after.txt)

PR #464, opened 2026-08-19
[cmake] Add `TEST_SUITE_PROFILE_DIR` to support out of build tree PGO Profile

mustartt:
Currently, PGO profiles produced/merged is written next to the test-suite executable, which forces the PGO collection/consumption through a single build directory. But it is often the case, we want to have multiple builds with varying compiler/flag combinations with the same set of profiles, which the current configuration does not allow. This is useful for experiments and CI pipeline.

So a new `TEST_SUITE_PROFILE_DIR=<dir>` is introduced to decouple the 2 stages. When set, the build tree will be mirrored under `<dir>` and the benchmark path will be used as the final merge pass by `llvm-profdata`. So an example use case would be as follows:

```
# profile generation
cmake -S <test-suite> -B build-gen -DTEST_SUITE_PROFILE_GENERATE=ON \
          -DTEST_SUITE_PROFILE_DIR=<custom profile directory> ... 
ninja -C build-gen
llvm-lit -sv build-gen

# profile use
cmake -S <test-suite> -B build-use1 -DTEST_SUITE_PROFILE_GENERATE=ON \
          -DTEST_SUITE_PROFILE_DIR=<custom profile directory> ... 
cmake -S <test-suite> -B build-use2 -DTEST_SUITE_PROFILE_GENERATE=ON \
          -DCMAKE_CXX_FLAGS="-mllvm -hot-callsite-threshold=1000" \
          -DTEST_SUITE_PROFILE_DIR=<custom profile directory> ...
          
<custom profile directory>:
  SingleSource/
    Benchmarks/
      McGill/
        chomp.profdata
        exptree.profdata
        misr.profdata
        queens.profdata
  ...
```


apolloww:
I don't have much context on this test suite, so I asked Claude for a review. It came back with an issue of "TEST_SUITE_PROFILE_USE + TEST_SUITE_PROFILE_DIR cannot configure when tools/test is in TEST_SUITE_SUBDIRS (it is, by default) -- even with a fully populated profile directory." I don't fully trust its assessment. Maybe you can check if this is really an issue. 

Below is the repro script (fill in your own paths to llvm-project and clnag/clang++/llvm-profdata)

```bash
# Shows that TEST_SUITE_PROFILE_USE + TEST_SUITE_PROFILE_DIR cannot configure
# when tools/test is in TEST_SUITE_SUBDIRS (it is, by default) -- even with a
# fully populated profile directory.
#
# Every build dir is created fresh under $WORK, so no stale CMakeCache can
# leak TEST_SUITE_PROFILE_DIR between cases (it is a sticky CACHE PATH entry).
set -u

SRC=${SRC:-<PATH_TO_LLVM_PROJECT>/llvm-project}
CC=$SRC/build/bin/clang
CXX=$SRC/build/bin/clang++
PROFDATA=$SRC/build/bin/llvm-profdata

WORK=/tmp/tsprof-repro
GEN=$WORK/gen; P=$WORK/profiles
CASE1=$WORK/case1-fail; CASE2=$WORK/case2-no-profdir; CASE3=$WORK/case3-feature-ok

cleanup() { echo; echo "### cleanup: rm -rf $WORK"; rm -rf "$WORK"; }
trap cleanup EXIT

rm -rf "$WORK"
mkdir -p "$GEN" "$P" "$CASE1" "$CASE2" "$CASE3"

cfg() { # cfg <builddir> <extra cmake args...>
  local d=$1; shift
  ( cd "$d" && cmake -G Ninja "$SRC" \
      -DCMAKE_C_COMPILER="$CC" -DCMAKE_CXX_COMPILER="$CXX" \
      -DCMAKE_BUILD_TYPE=Release "$@" > cfg.log 2>&1 )
}

# ---------------------------------------------------------------- setup ----
echo "### setup: profile-generate build + populate \$P"
cfg "$GEN" -DTEST_SUITE_SUBDIRS="tools/test;SingleSource/Benchmarks/Dhrystone" \
           -DTEST_SUITE_PROFILE_GENERATE=ON -DTEST_SUITE_PROFILE_DIR="$P" \
  || { echo "setup configure FAILED"; tail -20 "$GEN/cfg.log"; exit 1; }
( cd "$GEN" && ninja > build.log 2>&1 ) \
  || { echo "setup build FAILED"; tail -20 "$GEN/build.log"; exit 1; }

# llvm-lit is not installed here, so do what litsupport/modules/profilegen.py
# does: merge into $P/<exe path relative to build dir>.profdata
mkdir -p "$P/SingleSource/Benchmarks/Dhrystone"
for t in dry fldry; do
  ( cd "$GEN/SingleSource/Benchmarks/Dhrystone" \
    && LLVM_PROFILE_FILE="$WORK/$t.profraw" ./$t >/dev/null 2>&1
    "$PROFDATA" merge -output="$P/SingleSource/Benchmarks/Dhrystone/$t.profdata" "$WORK/$t.profraw" )
done
echo "    profile dir:"; find "$P" -type f | sed "s|$P|    \$P|"

# --------------------------------------------------------------- case 1 ----
echo
echo "### CASE 1 (expect FAIL): PROFILE_USE + PROFILE_DIR, subdirs include tools/test"
cfg "$CASE1" -DTEST_SUITE_SUBDIRS="tools/test;SingleSource/Benchmarks/Dhrystone" \
             -DTEST_SUITE_PROFILE_USE=ON -DTEST_SUITE_PROFILE_DIR="$P"
echo "    configure exit=$?"
grep -A3 "CMake Error" "$CASE1/cfg.log" | sed 's|^|    |'

# --------------------------------------------------------------- case 2 ----
echo
echo "### CASE 2 (expect PASS): same, but WITHOUT PROFILE_DIR -- pre-patch behaviour"
cfg "$CASE2" -DTEST_SUITE_SUBDIRS="tools/test;SingleSource/Benchmarks/Dhrystone" \
             -DTEST_SUITE_PROFILE_USE=ON
echo "    configure exit=$?  (profile-less targets now fail at compile time instead)"

# --------------------------------------------------------------- case 3 ----
echo
echo "### CASE 3 (expect PASS): PROFILE_USE + PROFILE_DIR, tools/test excluded"
cfg "$CASE3" -DTEST_SUITE_SUBDIRS="SingleSource/Benchmarks/Dhrystone" \
             -DTEST_SUITE_PROFILE_USE=ON -DTEST_SUITE_PROFILE_DIR="$P"
echo "    configure exit=$?"
( cd "$CASE3" && ninja > build.log 2>&1 ); echo "    build     exit=$?"
grep -o '\-fprofile-instr-use=[^ ]*' "$CASE3/build.ninja" | sort -u | sed 's|^|    |'

```

mustartt:
> "TEST_SUITE_PROFILE_USE + TEST_SUITE_PROFILE_DIR cannot configure when tools/test is in TEST_SUITE_SUBDIRS (it is, by default) -- even with a fully populated profile directory." I don't fully trust its assessment. Maybe you can check if this is really an issue.

@apolloww Thanks for catching that, the symptom is correct. But the underlying problem is that `tools/test`, i.e. the test harness binary `fpcmp`, `not`, `timeit`, etc, should not get PGO flags at all. 

So might as well fix that existing issue now.

CMakeLists.txt
```diff
@@ -149,6 +149,11 @@ endif()
 option(TEST_SUITE_USE_IR_PGO
   "Use IR PGO instrumentation (requires TEST_SUITE_PROFILE_GENERATE)" OFF)
 
+# Directory tree holding the `<target>.profdata` files written by
+# TEST_SUITE_PROFILE_GENERATE and read by TEST_SUITE_PROFILE_USE.
```
lukel97, line 153: Maybe document that this defaults to writing the profiles alongside the executables

cmake/modules/TestSuite.cmake
```diff
@@ -57,11 +58,25 @@ function(llvm_test_executable_no_test target)
...
+      file(RELATIVE_PATH target_reldir "${CMAKE_BINARY_DIR}" "${CMAKE_CURRENT_BINARY_DIR}")
+      set(profdata_file "${TEST_SUITE_PROFILE_DIR}/${target_reldir}/${target}.profdata")
+      if(NOT EXISTS "${profdata_file}")
+        message(FATAL_ERROR "No profile found for ${target}: ${profdata_file}")
+      endif()
+      append_target_flags(COMPILE_FLAGS ${target} -fprofile-instr-use=${profdata_file})
+      append_target_flags(LINK_LIBRARIES ${target} -fprofile-instr-use=${profdata_file})
+      get_target_property(_profdata_srcs ${target} SOURCES)
+      foreach(_src ${_profdata_srcs})
+        set_property(SOURCE ${_src} APPEND PROPERTY OBJECT_DEPENDS "${profdata_file}")
```
lukel97: Just checking this is so that if the profile changes the binaries get rebuilt?

litsupport/modules/profilegen.py
```diff
@@ -17,14 +19,25 @@ def _mutateScript(context, script):
     return testplan.mutateScript(context, script, _mutateCommandline)
 
 
+def _profdataPath(context):
+    """Return the path of the merged profile of the benchmark under test."""
+    relpath = os.path.splitext(os.path.join(*context.test.path_in_suite))[0]
```
lukel97, line 24: @copilot is `context.test.path_in_suite` the same as `context.executable`? Can these diverge in the default `if not profile_dir` path

cmake/modules/TestSuite.cmake
mustartt: Yes it does.

litsupport/modules/profilegen.py
mustartt: It does not diverge for the non `tools/test` binaries if `profile_dir` is not set. When `profile_dir` is set, it uses the lit `<target>.test` to generate the `.profdata` merge path. 

The `litsupport/shellcommand.py` is quite outdated https://github.com/llvm/llvm-test-suite/blob/824802c01e93a8d49a77384da4e68c76d1021953/litsupport/shellcommand.py#L146
so `context.executable` is populated by `shellcommand.getMainExecutable`, which is the first `RUN:` not in the ignore list. 

I think the divergence is fine with a `profile_dir` specified, and this was verified against SPEC and the test-suite.

CMakeLists.txt
mustartt: Good idea. Added it now.

lukel97 (approved):
LGTM

cmake/modules/TestSuite.cmake
mustartt: I'm realizing that this might not be the best idea, as it changes the existing behavior to now require the profdata to always exist. 
```
      if(NOT EXISTS "${profdata_file}")
        message(FATAL_ERROR "No profile found for ${target}: ${profdata_file}")
      endif()
```
which break the use case where people only run a subset of the test-suite. Let me back this change out.

mustartt:
@lukel97 Thanks for taking a look. 

I'm going to back out the `set_property OBJECT_DEPENDS` from this patch. As I'm just realizing that it breaks an existing common use case.

> I'm realizing that this might not be the best idea, as it changes the existing behavior to now require the profdata to always exist.

      if(NOT EXISTS "${profdata_file}")
        message(FATAL_ERROR "No profile found for ${target}: ${profdata_file}")
      endif()
> which break the use case where people only run a subset of the test-suite. 

Before this change, if the profile data didn't exist , uring the `PROFILE_USE` build we would get `clang: error: Error in reading profile` when the target is being built. Now, we would get an error at the CMake configure stage.

I will re-test with this bit backed out and attempt to land it.

PR #465, opened 2026-08-19
scope os.walk to CMakeFiles/<exe>.dir/ in non-SingleSource mode

kkwli:
_getCompileTime walked the test file's parent directory to find .o.time files. In non-SingleSource mode the filename prefix is empty, so the walk matched every .o.time file in the tree. When multiple tests share a build directory this caused two problems:

* every test accumulated .o.time files from sibling tests, producing identical fabricated compile_time values across unrelated tests

* the walk visited the entire test directory tree on every test, even for compile-only tests that produce no executable and no .o.time files at all

In non-SingleSource mode each test has its own CMakeFiles/<exe>.dir/ containing only that test's object-file timings. Scope the walk to that directory. If it does not exist (compile-only tests), return zeroed metrics immediately.

Assisted by: IBM Bob

### DIFF
--- a/litsupport/modules/compiletime.py
+++ b/litsupport/modules/compiletime.py
@@ -15,8 +15,23 @@ def _getCompileTime(context):
     link_time = 0.0
     compile_maxrss = 0
     link_maxrss = 0
-    dir = os.path.dirname(context.test.getFilePath())
-    for path, subdirs, files in os.walk(dir):
+    test_dir = os.path.dirname(context.test.getFilePath())
+    if context.config.single_source:
+        # SingleSource mode: multiple tests share a directory; the prefix
+        # already restricts matches to this test's files.
+        walk_dir = test_dir
+    else:
+        # Normal mode: each test has its own CMakeFiles/<exe>.dir/ subdirectory.
+        exe_name = os.path.basename(context.executable)
+        walk_dir = os.path.join(test_dir, "CMakeFiles", exe_name + ".dir")
+        if not os.path.isdir(walk_dir):
+            return {
+                "compile_time": compile_time,
+                "link_time": link_time,
+                "compile_maxrss": compile_maxrss,
+                "link_maxrss": link_maxrss,
+            }
+    for path, subdirs, files in os.walk(walk_dir):
         for file in files:
             if file.endswith(".o.time") and file.startswith(prefix):
                 fullpath = os.path.join(path, file)


kkwli:
Timing on AMD (8 workers)

|  | before | after |
|--| --|--|
| Fortran suite | 194.7 sec | 29 sec |

Timing on P9 (160 workers)
|  | before | after |
|--| --|--|
| Fortran suite | 41.8 sec | 13.8 sec |

litsupport/modules/compiletime.py
```diff
@@ -15,8 +15,23 @@ def _getCompileTime(context):
...
-    for path, subdirs, files in os.walk(dir):
+    test_dir = os.path.dirname(context.test.getFilePath())
+    if context.config.single_source:
+        # SingleSource mode: multiple tests share a directory; the prefix
+        # already restricts matches to this test's files.
+        walk_dir = test_dir
+    else:
+        # Normal mode: each test has its own CMakeFiles/<exe>.dir/ subdirectory.
+        exe_name = os.path.basename(context.executable)
+        walk_dir = os.path.join(test_dir, "CMakeFiles", exe_name + ".dir")
```
tarunprabhu: Looking for `exe_name + ".dir"` feels fragile. Presumably this is an internal convention used by `cmake`. While it is probably unlikely to change in practice, if it does, it would probably break this in a subtle way. What if we were to only look at `${test_dir}/CMakeFiles`? That still does seem like an internal cmake convention, too, though.

I don't have a strong opinion on this. What do you think?

tarunprabhu (commented):
Thanks. This looks to be really useful.

litsupport/modules/compiletime.py
kkwli: The `exe_name + ".dir"` is a cmake thing when doing `add_library()` or `add_executable`. I guess if somehow cmake changes the convention, there will be a lot of things broken. In addition, changing to `... = os.path.join(test_dir, "CMakeFiles")` will leave the inaccurate timing report issue there (bullet 1 in the description). 

Doing `${test_dir}/CMakeFiles` makes the speedup less than before but it still helps.

| | before | after (`os.path.join(test_dir, "CMakeFiles", exe_name + ".dir")` | after (`os.path.join(test_dir, "CMakeFiles")`|
|--|--|--|--|
| Fortran suite | 194.7 | 41.7 | 106 |

I suggest to keep `exe_name + ".dir"` but open to other opinions.

kkwli:
Ping

litsupport/modules/compiletime.py
tarunprabhu: Thanks for the clarification.

tarunprabhu (commented):
This LGTM. But I'll hold off on approving to give the others from whom I have requested reviews a chance to take a look.

litsupport/modules/compiletime.py
tomershafir: This would break link time metrics. Need 2 walk dirs: one for compile time metrics, and one for link time metrics

tomershafir (changes requested):
I think the direction of a focused collection is good, but we may be able to solve it differently. I think we can try persist compile time and link time files from build stage to run/llvm-lit stage in a file, and avoid a implicit dependency on CMake - like we do for timeit files.

Note that anyway such a change effectively alters the contract of llvm-test-suite, and may break external test suites (`-DTEST_SUITE_SUBDIRS`). I think it should be ok to extend the contract this way, given the tight coupling external test suites already have today with llvm-test-suite.

kkwli:
> I think the direction of a focused collection is good, but we may be able to solve it differently. I think we can try persist compile time and link time files from build stage to run/llvm-lit stage in a file, and avoid a implicit dependency on CMake - like we do for timeit files.
> 
> Note that anyway such a change effectively alters the contract of llvm-test-suite, and may break external test suites (`-DTEST_SUITE_SUBDIRS`). I think it should be ok to extend the contract this way, given the tight coupling external test suites already have today with llvm-test-suite.

Thanks a lot @tomershafir. I take your suggestion to write the compile and link time into a summary file at build time. So that `compiletime.py` does not need to walk the directories. However, the dependency of the `CMakeFiles` directory is still there. I insert a TODO in `aggregate-compile-time.py` for further improvement when the minimum version of cmake gets bumped up.

I compared between the original and the current version. For Fortran suite (~6900 tests), the build time is within 3%  and the run time maintains around 2x speedup on Linux PPC; the build time more or less the same and the run time ~7x speedup on Linux x86.

tomershafir:
The direction I tried to suggest is a bit different.

For example, we have an internal test suite integrated with llvm-test-suite that is not using `add_executable`/`add_library` of CMake and thus lacks a `CMakeFiles/.dir` path. I dont think making it write a CMake path is a good idea.

I think we need to create a generic contract for sharing the location of compile-time and link-time files, such that each suite is responsible to follow the contract. Because we have 2 distinct processes (build and llvm-lit) we should use a file for that (unlike runtime which is passed in-memory via `context.timefiles`). The available handle which is unambiguous at test time is `context.test.getFilePath()`, so maybe we can generate co-located metadata files at build time which contains a direct list of compile/link files, and are named like `f"{context.test.getFilePath()}.o.time.list"` and `f"{context.test.getFilePath()}.link.time.list"`.

This way we may be able to avoid: overcount, undercount, globbing, support external test suites and not break them.

I admit it does add complexity but I think it can provide what we need.

kkwli:
@tomershafir I made the requested change. Please let me know if it aligns with what you envision. Thanks.

kkwli:
Ping

PR #466, opened 2026-08-26
[Fortran] Add LAPACK as an external package

DavidTruby:
LAPACK is one of the most widely used Fortan libraries and has an extensive suite of numerical precision tests. Allowing this to be easily tested as part of the llvm test-suite would benefit flang testing.


External/LAPACK/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+find_package(Python3 REQUIRED COMPONENTS Interpreter)
+
+set(LAPACK_BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/lapack-build")
+
+set(LAPACK_CMAKE_ARGS
+  "-DCMAKE_BUILD_TYPE:STRING=${CMAKE_BUILD_TYPE}"
+  "-DCMAKE_C_COMPILER:FILEPATH=${CMAKE_C_COMPILER}"
+  "-DCMAKE_Fortran_COMPILER:FILEPATH=${CMAKE_Fortran_COMPILER}"
+  "-DCMAKE_C_FLAGS:STRING=${CMAKE_C_FLAGS}"
+  "-DCMAKE_Fortran_FLAGS:STRING=${CMAKE_Fortran_FLAGS} -fstack-arrays -ffp-contract=off"
```
tarunprabhu, line 26: Is it necessary to hard-code `-fstack-arrays` and `-ffp-contract=off`?

If they were to appear before `${CMAKE_Fortran_FLAGS}`, would that allow them to be overridden by the configuration options used to build the test suite? I can imagine someone intentionally configuring the test suite with `-fno-stack-arrays` and/or `-ffp-contract=fast`, for example.

tarunprabhu (commented):
This looks reasonable, but I haven't tried it out. Unfortunately, I might not have time to do this for the next few weeks.

External/LAPACK/CMakeLists.txt
DavidTruby: LAPACK's tests involve a lot of numerical accuracy tests which require `-ffp-contract=off` at least. In my testing the tests fail with `-ffp-contract=fast` so we do need to override that, unless we specifically want to allow the tests to fail if someone explicitly passes -ffp-contract=fast.
I think the tests would still pass with `-fno-stack-arrays` but LAPACK throws a lot of warnings about it which is why I added it.

tarunprabhu: It's probably worth leaving a comment explaining this when setting up the configuration parameters. While I generally prefer giving users a choice where possible, in this case, a failure in LAPACK's tests has to do with the floating point accuracy required by LAPACK - they do not reflect a failure of the compiler. We could consider adding a warning in case users add `-ffp-contract=fast` to `CMAKE_Fortran_FLAGS`, but it's probably not worth the extra effort either. 

Thanks for the clarification.

tarunprabhu (approved):
This LGTM, but I haven't had a chance to test it. won't be able to get to it for another several weeks at least, so I am approving this so as not to block it.

tblah (approved):
Works for me on an aarch64 gnu+linux machine

PR #469, opened 2026-09-02
[HIP] After CMake bump, passing -xhip is required

jplehr:
In https://github.com/llvm/llvm-test-suite/pull/468 the CMake version minimum was raised to 3.20.0
This led to the policy CMP0119 changing meaning.

> In CMake 3.19 and below, setting this property causes CMake to
> compile the source file using the compiler for the specified language.
> ...
> CMake 3.20 and above prefer to also explicitly tell the compiler to use
> the specified language using a flag such as -x c on all compilers for
> which such flags are known.

This turned our buildbot https://lab.llvm.org/buildbot/#/builders/232 red as it started to build HIP files as plain C++.

With this patch, we add an explicit -xhip for all such files to the end of the command line, to overrule previously set -xc++.

Setting the CMake language to HIP instead of CXX would have introduced some linker problems in the specific case of the HIP test suite. In the test suite, we have the ability to test various locally installed ROCm versions. Setting the CMake language to HIP, however, it would only ever link one dedicated toolchain. This would be undesirable for that particular test suite.

Assisted-by: Claude


tomershafir (approved):

Meinersbur (approved):
CMake has first class CUDA/HIP support: https://cmake.org/cmake/help/latest/command/enable_language.html, so this could be
```
set_source_files_properties(${File} PROPERTIES
        LANGUAGE HIP)
```

A problem with the `-xhip` option is that it is order-dependent, and CMake may not guarantee a specific order (CMake's `-x c++` might be inserted after the custom `-xhip`).

On the other side, `enable_language(CUDA HIP)` will trigger a lot of other complication, such that the `CMAKE_HIP/CUDA_COMPILER` is working. That's probly not worth it, but I would enjoy a TODO/FIXME comment about this.

cmake/modules/GPUTestVariant.cmake
```diff
@@ -15,11 +15,19 @@ endmacro (get_version)
...
+# CUDA/HIP toolchain and leaves the offload headers unreachable. Re-assert the
+# offload language after it; the last -x on the command line wins.
 macro(gpu_glob Var)
   file(GLOB FileList ${ARGN})
   foreach(File IN LISTS FileList)
-    if(${File} MATCHES ".*\.cu$" OR ${File} MATCHES ".*\.hip$")
-      set_source_files_properties(${File} PROPERTIES LANGUAGE CXX)
+    if(${File} MATCHES ".*\.cu$")
+      set_source_files_properties(${File} PROPERTIES
+        LANGUAGE CXX COMPILE_OPTIONS "-xcuda")
```
Meinersbur: Prefer `set_property(SOURCE ${File} APPEND PROPERTY COMPILE_OPTIONS "-xhip")` to append, not override previously set compile options

jplehr:
> CMake has first class CUDA/HIP support: https://cmake.org/cmake/help/latest/command/enable_language.html, so this could be
> 
> ```
> set_source_files_properties(${File} PROPERTIES
>         LANGUAGE HIP)
> ```
> 
> A problem with the `-xhip` option is that it is order-dependent, and CMake may not guarantee a specific order (CMake's `-x c++` might be inserted after the custom `-xhip`).
> 
> On the other side, `enable_language(CUDA HIP)` will trigger a lot of other complication, such that the `CMAKE_HIP/CUDA_COMPILER` is working. That's probly not worth it, but I would enjoy a TODO/FIXME comment about this.

From what I found with this is that this will configure-time use one specific HIP toolchain. But the test suite can run all the tests suites against multiple locally installed ROCm versions. With the HIP support from CMake, my understanding is that this will go away.

I have, however, put on my list to look more into it because I don't like the current CMake too much anyway.

cmake/modules/GPUTestVariant.cmake
jplehr: For some cases in the test suite this will result in several `-xhip` additional parameters to the command line, given how that's constructed.
It is ugly, but not harmful. What's your preference?

External/HIP/CMakeLists.txt
```diff
@@ -93,7 +93,9 @@ macro(create_one_hipify_cuda_test TestName TestSource VairantOffload VariantSuff
   set(_hip_src "${TestName}.hip")
   set(_hipify_target "${TestName}-hipify")
 
-  set_source_files_properties(${_hip_src} PROPERTIES LANGUAGE CXX)
+  # -xhip must follow the -x c++ that CMP0119 injects alongside LANGUAGE; see gpu_glob.
+  set_source_files_properties(${_hip_src} PROPERTIES
+    LANGUAGE CXX COMPILE_OPTIONS "-xhip")
```
jmmartinez: Maybe a dumb question, but why do we set CXX instead of HIP as the language?

jmmartinez (commented):
Minor comment otherwise it looks good.

External/HIP/CMakeLists.txt
jplehr: Not a dumb question, and it was my preferred solution, but: https://github.com/llvm/llvm-test-suite/pull/469#issuecomment-5508325244

jplehr:
I have changed the implementation a bit to make the `-xhip` a target setting instead of a file setting to work around the point of it being added several times for various test cases.

Kewen12 (approved):

jplehr:
I'm going to land with these recent changes. If they are undesirable, please ping me here and I can do the implementation differently.

PR #470, opened 2026-09-02
[Bitcode] Pass -xir explicitly after CMake bump

tomershafir:
The minimum version of CMake has been bumped in https://github.com/llvm/llvm-test-suite/pull/468. It changed the way CMake handles the `LANGUAGE` property - it now defines the language at which the input files are written, and not just the compiler to use which would later infer the language. For example in this case `LANGUAGE CXX` injects `-x cxx` and fails to build the bitcode files. Same for C.

This patch overrides the injected options with an ad-hoc `-xir`.


wangpc-pp (approved):
LGTM.

And just out of curious, does cmake support something like `LANGUAGE LLVMIR`? If so, that will make life easier?

Bitcode/Benchmarks/Halide/bilateral_grid/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 file(GLOB bcsources ${CMAKE_CURRENT_SOURCE_DIR}/../common/x86_halide_runtime.bc ${CMAKE_CURRENT_SOURCE_DIR}/bilateral_grid.bc)
-SET_SOURCE_FILES_PROPERTIES(${bcsources} PROPERTIES LANGUAGE CXX)
+# -xir must follow the -x c++ that CMP0119 injects alongside LANGUAGE CXX.
+SET_SOURCE_FILES_PROPERTIES(${bcsources} PROPERTIES LANGUAGE CXX COMPILE_OPTIONS "-xir")
```
Meinersbur: Prefer `set_property(SOURCE ${bcsources} APPEND PROPERTY COMPILE_OPTIONS "-xir")` to append, not override previously set compile options

tomershafir:
It doesnt seem like CMake supports LLVM IR: https://cmake.org/cmake/help/latest/command/enable_language.html

Meinersbur (approved):

PR #471, opened 2026-09-07
[Fortran] Integrate RIKEN F2008 conformance tests into CMake/lit build

h-murai:
## Summary
Wire `Fortran/riken` into the test-suite's normal CMake + lit build, the
same way as `Fortran/cray` and `Fortran/UnitTests`: each `_NNN` test
source is compiled at build time with `CMAKE_Fortran_COMPILER`, run at
test time, and its combined stdout/stderr (+ exit status) is checked
against a `.reference_output` file.

- 176 of 194 normal-case (`_NNN`) sources build, run, and verify against
  a real flang build; 18 are excluded/gated with documented reasons
  (confirmed flang bugs worth an upstream report, macOS-only libm gaps,
  missing `REAL128` support, missing fixtures, or checks that need
  custom verify logic beyond `llvm_singlesource()`). Full breakdown in
  `Fortran/riken/README.md`.
- Error-case (`_cNNN`) sources (220 files) are filtered out of the
  build for now; they expect a compile-time diagnostic and need a
  dedicated "expect failure" CMake helper. The README documents prior
  art from `Fortran/gfortran`'s own 2023 design discussion on this
  exact question
  ([discourse.llvm.org/t/add-gfortran-tests-to-llvm-test-suite/69408](https://discourse.llvm.org/t/add-gfortran-tests-to-llvm-test-suite/69408))
  and the recommended follow-up approach (reuse
  `gfortran_add_compile_test()`'s pass/fail-only verification).

See `Fortran/riken/README.md` for the full status table, per-test
exclusion rationale, and the naming/layout conventions used throughout
this subtree.

## Test plan
- [x] Configured with `-DTEST_SUITE_SUBDIRS=Fortran -DCMAKE_Fortran_COMPILER=flang`
  and built with `ninja`.
- [x] Ran `llvm-lit Fortran/riken` end-to-end against a real flang
  build (Homebrew flang 22.1.8, arm64 macOS); all enabled `_NNN` tests
  pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tarunprabhu:
Thank you very much for contributing these. I won't have time to take a look at this for the next two weeks, but a quick glance raised some questions:

- The gfortran tests are generally categorized as "compile" and "execute" tests. The former checks that the compiler correctly accepts/rejects certain code. The latter checks the runtime behavior of the compiled executable. Are the tests in this suite exclusively "compile", exclusively "execute" or a mix of both?

- A quick glance at several tests showed them to be "lit" tests. Each test is `_compiled_` at test time. This is not how most of the llvm-test-suite works. The gfortran tests do not do this either. Instead, the gfortran tests build all tests at test-suite build time. At runtime, the "execute" tests are actually executed while the output of the "compile" tests are examined for the expected error (or lack of error as the case may be). Is it possible for these tests to conform to that behavior?

- In several tests, I noticed invocations of both `flang` and `gfortran`. We generally should not be testing `gfortran` when running the LLVM test suite.

h-murai:
@tarunprabhu Thank you for reviewing this PR.

> * The gfortran tests are generally categorized as "compile" and "execute" tests. The former checks that the compiler correctly accepts/rejects certain code. The latter checks the runtime behavior of the compiled executable. Are the tests in this suite exclusively "compile", exclusively "execute" or a mix of both?

It's a mix, by RIKEN's own naming convention:

* `_NNN` files (e.g. `25_blockConstruct_001.f90`) are execute tests: build, run, and compare combined stdout/stderr + exit status against a `.reference_output`. 176 of 194 are wired into the build in this PR (the remaining 18 are excluded/gated with reasons documented in Fortran/riken/README.md).
* `_cNNN` files (e.g. `10_pointerInitialization_c004.f90`) are compile tests: they're expected to fail to compile with a diagnostic. There are 220 of these, and they are not wired into the build in this PR, since `llvm_singlesource()` only supports "build successfully and run" targets today. Follow-up plan is written up in the README under "Error-case (`_cNNN`) tests: prior art and future plan".

> * A quick glance at several tests showed them to be "lit" tests. Each test is _compiled_ at test time. This is not how most of the llvm-test-suite works. The gfortran tests do not do this either. Instead, the gfortran tests build all tests at test-suite build time. At runtime, the "execute" tests are actually executed while the output of the "compile" tests are examined for the expected error (or lack of error as the case may be). Is it possible for these tests to conform to that behavior?

Yes, and this PR already does that for the execute (`_NNN`) tests. Each of the 176 wired-in sources becomes a real CMake target via `llvm_singlesource()` (cmake/modules/SingleMultiSource.cmake), so it's compiled at `ninja`/build time, same as `Fortran/gfortran`. At lit-run time we just execute the already-built binary and diff its output against `.reference_output` — no compilation happens during `llvm-lit`.

The original RIKEN-supplied `! RUN: %flang %s -o %t && %t | FileCheck %s`-style lit directives have already been stripped from all 176 wired-in sources (see e.g. `25_blockConstruct_001.f90`), and in any case the top-level `lit.cfg` sets `config.suffixes = [".test"]`, so lit never parses `! RUN:` lines inside `.f90` files directly — only the CMake-generated `.test` files are lit inputs.

What you likely spotted is that 16 of the 18 currently-excluded `_NNN` files, plus all 220 not-yet-wired `_cNNN` files, still carry those original `! RUN:` lines as inert leftover text (never touched by CMake or lit). I can strip those from the 16 excluded `_NNN` files in this PR for consistency; the `_cNNN` ones will get cleaned up together with the compile-test follow-up work.

> * In several tests, I noticed invocations of both flang and gfortran. We generally should not be testing gfortran when running the LLVM test suite.

No `gfortran` invocation actually happens in this build/test flow. As noted above, `lit.cfg`'s `config.suffixes = [".test"]` means lit never interprets `! RUN:` lines inside `.f90` sources — it only runs the generated `.test` files, which only ever invoke the configured `CMAKE_Fortran_COMPILER` (flang).

That said, the concern is valid as a source-hygiene issue: those `! RUN: %gfortran ...` lines are leftover from RIKEN's original distribution and are currently dead comments in 16 excluded `_NNN` files and all 220 `_cNNN` files. I'll strip them from the 16 excluded `_NNN` files in this PR; the `_cNNN` ones are already flagged for the same treatment when that follow-up work lands.

PR #472, opened 2026-09-07
[HIP] Propagate image comparison result as exit code

jplehr:
In the amdgpu-hip-blender buildbot is an issue with the image comparison return code: it is ignored. This patch fixes it and surfaces potentially miscomparing images. Thank you @kasaurov for catching this.

It actually exposed an error that (probably I) introduced: The current buildbot setup does compare two different frames of a blender scene. This leads to large image discrepancy metrics and would turn the buildbot red when we land this change.

Hence, we need to update the buildbot image itself to contain the correct comparison images/frames. The change is queued up internally for review. Once that is done, we can update the buildbot image and land this change.

The change:
compare_image.py computed a -1/0 result, printed it and wrote the diff image, but main() returned without ever calling sys.exit(). The process therefore always exited 0, so compare_return_code in test_blender.sh was always 0 and an image mismatch could not fail the Blender test.

Observed on an AMD HIP buildbot: a scene rendering at SSIM 0.033 and MSE 1.8e+04, against thresholds of 0.9 and 1000, still produced "Blender test passes."

Assisted-by: Claude Code


yxsamliu (approved):
LGTM.

PR #474, opened 2026-09-09
[HIP] Gate the hipified CUDA tests behind ENABLE_HIP_HIPIFY_TESTS

jplehr:
The hipified External/CUDA tests were built whenever hipify-perl happened
to be found and were dropped with only a warning when it was not, so a
configuration missing hipify-perl reported the same success as one that
ran all five test groups.

Give the group its own ENABLE_HIP_HIPIFY_TESTS cache variable, defaulting
to ON to keep the current intent, and turn the missing-hipify-perl
warning into an error that names the flag to disable. The tests now build
and run through their own hip-tests-hipify / check-hip-hipify targets
instead of being folded into the simple ones, so a bot can ask for
exactly the group it wants.

This is the first of several groups to be separated this way. To support
that, hip-tests-all is assembled from a list rather than an if/else
chain, and the missing check-hip-all counterpart is added.

Note that check-hip-simple no longer covers the hipified tests; use
check-hip-all for the previous coverage.

Assisted-by: Claude Code

---


yxsamliu (approved):
LGTM

jmmartinez (approved):
> A group that is enabled validates its prerequisites while configuring and fails there if they are missing, instead of quietly contributing no tests

Love this

PR #475, opened 2026-09-09
[HIP] Declare ENABLE_HIP_CATCH_TESTS with the other test group switches

jplehr:
The Catch group already had the shape the other groups are being moved
to: an opt-in cache variable, prerequisite validation that fails during
configuration, and its own hip-tests-catch / check-hip-catch pair. Only
the declaration sat apart, inside HipCatchTests.cmake.

Move it next to ENABLE_HIP_HIPIFY_TESTS so every group switch is declared
in one place, and use option() to match. The default stays OFF and the
group-specific tuning variables stay with the module that reads them.

No functional change.

Assisted-by: Claude Code

---


yxsamliu (approved):
LGTM

External/HIP/HipCatchTests.cmake
```diff
@@ -3,8 +3,10 @@
 
 include(CMakeParseArguments)
 
-# Global variables for Catch test configuration
-set(ENABLE_HIP_CATCH_TESTS OFF CACHE BOOL "Enable HIP Catch test framework and all catch test targets")
+# Global variables for Catch test configuration.
+# ENABLE_HIP_CATCH_TESTS itself is declared alongside the other HIP test group
```
jmmartinez, line 7: Nitpick, I'd not even mention `ENABLE_HIP_CATCH_TESTS` in here. I'd assume it from the context (since it's where all other test group switches are). This also avoids the message getting stale if we ever change the code structure again.

```suggestion
```

jmmartinez (approved):

DavidSpickett:
Possibly wrong PR but this one came in around the time. We have problems on our llvm-test-suite bots for example https://lab.llvm.org/buildbot/#/builders/17/builds/17448.

```
-- TEST_SUITE_HIP_ROOT: 
-- TEST_SUITE_HIP_MANAGED_DIR: /managed
-- HIP_CLANG_PATH: /home/tcwg-buildbot/worker/clang-aarch64-sve-vla/stage1.install/bin
-- Catch2 >= 3.8.1 not found on system, fetching v3.8.1...
<...>
  External/HIP/HipCatchTests.cmake:58 (FetchContent_Declare)
  External/HIP/CMakeLists.txt:38 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Error at catch2-populate-prefix/tmp/catch2-populate-mkdirs.cmake:6 (file):
  file failed to create directory:

    /managed/catch2/src

  because: No such file or directory
  /usr/local/share/cmake-3.28/Modules/ExternalProject.cmake:1790 (include)
  /usr/local/share/cmake-3.28/Modules/ExternalProject.cmake:4363 (_ep_set_directories)
  CMakeLists.txt:21 (ExternalProject_Add)


-- Configuring incomplete, errors occurred!

CMake Error at /usr/local/share/cmake-3.28/Modules/FetchContent.cmake:1667 (message):
  CMake step for catch2 failed: 1
  /usr/local/share/cmake-3.28/Modules/FetchContent.cmake:1819:EVAL:2 (__FetchContent_directPopulate)
  /usr/local/share/cmake-3.28/Modules/FetchContent.cmake:1819 (cmake_language)
  /usr/local/share/cmake-3.28/Modules/FetchContent.cmake:2033 (FetchContent_Populate)
  External/HIP/HipCatchTests.cmake:65 (FetchContent_MakeAvailable)
  External/HIP/CMakeLists.txt:38 (include)
```
I think HIP tests should be OFF for us and this PR looks like it's doing that.

Is this result expected? The test suite dir is cleaned every run, so it shouldn't be using cached values.

jplehr:
> Is this result expected? 

I don't think it is and is potentially rather related to https://github.com/llvm/llvm-test-suite/pull/484.

Let me take a look.

DavidSpickett:
In the prior build we did download catch but presumably to another location:
```
-- Adding directory External
-- TEST_SUITE_HIP_ROOT: 
-- TEST_SUITE_HIP_MANAGED_DIR: /managed
-- HIP_CLANG_PATH: /home/tcwg-buildbot/worker/clang-aarch64-sve-vla/stage1.install/bin
-- Catch2 >= 2.13.10 not found on system, fetching v2.13.10...
```
The last green build was 4b2c892865cbbb05ee996c88794a58e17d3adb81, and the first red was b6e040aae05c5c12f06f6029c4a5fec75d365e40.

jplehr:
Yes, I think the issue is that we used to download those into the build directory directly. However, given that we do not want to re-download all of these dependencies every time, we have introduced a shared place where such artifacts are placed. Basically to have them survive. Catch2 is one of the smaller parts of this.

It seems that we have not properly guarded this, I'm currently testing a quick fix locally.

PR #476, opened 2026-09-09
[HIP][NFC] Split create_local_hip_tests into one macro per test group

jplehr:
The macro built three unrelated groups of tests in one body: the in-tree
sources, the hipified External/CUDA sources, and the Blender rendering
test. Gating them individually means wrapping large blocks in place,
which obscures the change.

Extract create_simple_hip_tests, create_hipify_hip_tests and
create_blender_hip_test, and reduce create_local_hip_tests to dispatching
between them in the existing order. No behaviour change: the generated
target list and the test lists handed to llvm-lit are identical before
and after.

They stay macros because each appends to the VARIANT_*_TEST_TARGETS lists
that create_hip_test() passes to llvm-lit, which relies on running in the
caller's scope. Noted as a TODO for when the minimum CMake version
reaches 3.25 and block(PROPAGATE ...) can express that explicitly.

Assisted-by: Claude Code

---


yxsamliu:
Should this say `return(PROPAGATE ...)` instead? `block(PROPAGATE ...)` only propagates to the function scope, not to the function caller.

yxsamliu (approved):
LGTM

jmmartinez (approved):

PR #477, opened 2026-09-09
[HIP] Gate the in-tree HIP tests behind ENABLE_HIP_SIMPLE_TESTS

jplehr:
Give the in-tree External/HIP tests the same treatment as the hipified
and Catch groups: an opt-in cache variable, defaulting to ON, that also
controls whether the hip-tests-simple / check-hip-simple targets exist at
all. A consumer that only wants one other group no longer has to build
this one.

The Blender test is gated along with the group it currently belongs to,
so that turning the group off does not leave it without a target to run
in. The next commit gives it a switch of its own.

Assisted-by: Claude Code

---


yxsamliu:
When `ENABLE_HIP_SIMPLE_TESTS` is OFF, Blender is still created because `HIP_ENABLE_BLENDER_TEST` defaults to ON, but there is no target that runs it. Could we keep Blender under the simple-test condition in this commit and separate it in the next PR?

jplehr:
Good point. Thank you

yxsamliu (approved):
LGTM

PR #478, opened 2026-09-09
[HIP] Gate the Blender test behind ENABLE_HIP_BLENDER_TESTS

jplehr:
The Blender test was configured unconditionally and appended to the
simple group, so whether it ran was decided at run time by
test_blender.sh looking for a Blender install and a scene list. When
either was missing the script printed a skip message that
verify_blender.sh then rewrote into a pass, meaning an unstaged payload
was reported as a passing test.

Make the group opt-in, defaulting to OFF because its payload has to be
staged on the machine, and check for the Blender executable and the scene
list while configuring so an enabled group that cannot run fails there
instead. These are the same two things test_blender.sh checks; both can
still be redirected at run time via HIP_BLENDER_TEST_BIN_DIR and
HIP_BLENDER_TEST_SCENES_DIR, so only the default layout is validated.

The test moves out of the simple group into its own check-hip-blender
target. It only drives an already-installed Blender, so there is nothing
to build and it gets no hip-tests-blender counterpart.

Also add HIP_BLENDER_ALWAYS_SHOW_LOG, which prints the render log and the
scene summary even when the test passes. That needs both halves:
verify_blender.sh has to emit the log, and lit has to be asked for the
output of passing tests with -a. It can be flipped per run through the
environment variable of the same name.

Assisted-by: Claude Code

---


yxsamliu:
The runtime script supports custom Blender and scene paths through environment variables, but configuration checks only the default paths. This prevents valid runtime overrides when the default files are absent. Could we use configurable CMake paths here or leave the check until runtime?

yxsamliu:
The verifier still treats the missing Blender or scenes message as success. A bad runtime path can therefore pass without running the test. Since this group is explicitly enabled, should a missing runtime payload fail?

jplehr:
> The runtime script supports custom Blender and scene paths through environment variables, but configuration checks only the default paths. This prevents valid runtime overrides when the default files are absent. Could we use configurable CMake paths here or leave the check until runtime?

Yeah, makes sense. I think the important piece is that we fail loudly when blender tests are meant to be run but the toolchains are not there.
I personally think this should be configure-time variables.

jplehr:
> The verifier still treats the missing Blender or scenes message as success. A bad runtime path can therefore pass without running the test. Since this group is explicitly enabled, should a missing runtime payload fail?

I can add that to give a better safety net for unhappy paths.

yxsamliu:
@jplehr Re: missing Blender payload

Thanks for adding the runtime override. `verify_blender.sh.in` still exits with success when it sees the skip message, so a missing Blender binary or scene list can still pass the test. Could this path return a failure instead?

jplehr:
> @jplehr Re: missing Blender payload
> 
> Thanks for adding the runtime override. `verify_blender.sh.in` still exits with success when it sees the skip message, so a missing Blender binary or scene list can still pass the test. Could this path return a failure instead?

Thank you for catching this. I was so focused on the CMake changes that I forgot about this.

yxsamliu (approved):
LGTM

PR #479, opened 2026-09-09
[HIP] Name the external HIP suites like the other test groups

jplehr:
Kokkos, Ginkgo and rocPRIM were selected through EXTERNAL_HIP_TESTS_*
variables that were never declared, so they did not show up in cmake
-LAH, ccmake or cmake-gui and could only be found by reading this file.
They also did not match the ENABLE_HIP_*_TESTS scheme the other groups
now use.

Declare ENABLE_HIP_KOKKOS_TESTS, ENABLE_HIP_GINKGO_TESTS and
ENABLE_HIP_ROCPRIM_TESTS and forward them to the existing variables, so
callers passing the old spellings keep working unchanged. These suites
are fetched and built on the side and keep their own build-<name> /
test-<name> targets rather than joining the hip-tests-* / check-hip-*
pairs.

Assisted-by: Claude Code

---


yxsamliu (approved):
LGTM

jmmartinez (approved):

PR #480, opened 2026-09-09
[HIP] Remove old option spelling

jplehr:
This removes the old option spelling in favor of the new one which is
cleaner and more similar across the different test options.

---


yxsamliu:
The llvm-zorg HIP TPL builder still uses `EXTERNAL_HIP_TESTS_KOKKOS=ON`. Removing this option means `build-kokkos` is not created and the builder will fail. Could we migrate llvm-zorg first or keep the old aliases for now?

jplehr:
This is the last part of this stack. We will migrate the builder to the new spelling before this one lands.

jmmartinez (approved):

PR #482, opened 2026-09-10
[HIP] Fix generic-target helper lookup in catch tests

kasaurov:
Commit [2b51e7a7a ](https://github.com/llvm/llvm-test-suite/commit/2b51e7a7aa2653960774eb0251fc896d79e8cc2d) appended `-${VARIANT_SUFFIX}` to the generic-target helper executables so that two ROCm variants would stop declaring rules that generate the same output path. That fixed configure but broke the test at runtime: `hipSquareGenericTarget.cc` spawns the helpers by a hardcoded, unsuffixed name, `chmod` fails, so the shell never runs the helper, and `REQUIRE(std::system(cmd) == 0)` fails.

Keep the on-disk basename and make the directory per-variant instead. The helpers are now built into `catch_tests/generic-${VARIANT_SUFFIX}/`, and the main test runs from there via `WORKDIR`, reaching its own executable one level up. This also repairs Windows, where the suffix previously landed after `.exe` and produced `hipSquareGenericTargetOnly.exe-<variant>`.

No test source is touched, so the port stays aligned with upstream.


jplehr (approved):
thanks for fixing this!

PR #483, opened 2026-09-10
[Fortran][gfortran] Disable pointer_check_12.f90 test

DavidSpickett:
This test is supposed to use -fcheck=all to catch a pointer mistake before it becomes a segfault:
```
$ gfortran /tmp/test.f90 -o /tmp/test.o -O3 -fcheck=all $ /tmp/test.o
At line 13 of file /tmp/test.f90
Fortran runtime error: Pointer actual argument 'p' is not associated
```
This test was being run without the flag and expected to fail. Which it does as long as you're using -O0:
```
$ ~/llvm.install/bin/flang /tmp/test.f90 -o /tmp/test.o -O0 $ /tmp/test.o
Segmentation fault         (core dumped) /tmp/test.o
```
If you use -O3, it never completes. If you break in a debugger you'll see that it's calling `_FortranAProgramEndStatement` over and over. This happens because there is UB in the program which causes llvm to compile `main` to:
```
00000000000022ec <main>:
    22ec:       a9bf7bfd        stp     x29, x30, [sp, #-16]!
    22f0:       910003fd        mov     x29, sp
    22f4:       aa1f03e3        mov     x3, xzr
    22f8:       94000001        bl      22fc <_FortranAProgramStart>

00000000000022fc <_FortranAProgramStart>:
    22fc:       a9bf7bfd        stp     x29, x30, [sp, #-16]!
    2300:       910003fd        mov     x29, sp
    2304:       b00005e8        adrp    x8, bf000 <__abi_tag+0x168e8>
    2308:       aa0303e4        mov     x4, x3
```
So `_FortranAProgramStart` returns to itself.

I'm disabling this test for all the architectures I saw in the overrides file, as I think this is a generic problem.

This Compiler Explorer example shows the results: https://godbolt.org/z/r3PsMKf9Y

Somehow this test is running on Linaro's existing bots that do use -O3. The only explanation I have for that is that maybe the binary layout is different there and this causes a crash another way. A bad pointer, or stack exhaustion.

Either way, this test contains undefined behaviour so I don't think it's a good idea to be running it.


DavidSpickett:
I just listed all the seemingly common targets. If there is a bigger "disable everywhere" or "has UB don't run this" flag, I can set that instead.

Fortran/gfortran/regression/override.yaml
```diff
@@ -95,6 +95,12 @@
 "nint_2.f90":
   disabled_on: ["powerpc-*-aix"]
 
+# Is expected to be compiled with -fcheck=all, which we do not support yet.
+# Can be run at -O0 without -fcheck=all, but at -O1 and above it produces
+# an infinite loop calling _FortranAProgramEndStatement.
+"pointer_check_12.f90":
+  disabled_on: ["aarch64-*-*", "loongarch64-*-*", "ppc64le-*-*", "powerpc-*-*"]
```
tblah: I wonder if we should disable this on all architectures? IIUC the test is creating undefined behaviour. Even if it doesn't trip up other platforms currently it could do in the future or with different options.

tblah (commented):
Thanks for the patch

tblah:
> I just listed all the seemingly common targets. If there is a bigger "disable everywhere" or "has UB don't run this" flag, I can set that instead.

See Fortran/gfortran/regression/DisabledFiles.cmake. This probably belongs around pointer_check_10.f90 and pointer_check_11.f90 (roughly line 1011). I wonder if some of the other pointer_check tests should be disabled too?

Fortran/gfortran/regression/override.yaml
DavidSpickett: The only reason I didn't use ".*" was I didn't see it being done anywhere else. I've disabled it in DisabledFiles.cmake now anyway.

DavidSpickett:
> I wonder if some of the other pointer_check tests should be disabled too?

I'll check the rest today.

DavidSpickett:
| Test Name            | Annotations                    | Notes |
| -------------------- | ------------------------------ | ----- |
| pointer_check_1.f90  | xfailed                        | loop until crash |
| pointer_check_2.f90  | xfailed                        | loop until crash |
| pointer_check_3.f90  | xfailed                        | loop until crash |
| pointer_check_4.f90  | xfailed                        | loop until crash |
| pointer_check_5.f90  | already disabled, also xfailed | Fortran runtime error, core dump |
| pointer_check_6.f90  | xfailed                        | completes, because an access to `b` in `sub3` would be UB, but there isn't any |
| pointer_check_7.f90  | marked compile only            | compile only |
| pointer_check_8.f90  | marked compile only            | compile only |
| pointer_check_9.f90  |                                | completes, is valid as f2008 but not for 95 or 2003, which Flang does not support |
| pointer_check_10.f90 | disabled, also xfailed         | completes, expects an error when compiled with -std=f2003 which Flang does not support |
| pointer_check_11.f90 | disabled, also xfailed         | core dump |
| pointer_check_12.f90 | will disable, also xfailed     | core dump |
| pointer_check_13.f90 | marked compile only            | compile only |
| pointer_check_14.f90 |                                | completes, is expecting *not* to get a pointer check error. |

I think:
* 1/2/3/4 should be disabled rather than run, because they might manage to not crash one day.
* The tests that are xfail and disable should be only disabled.

Would you be ok with me landing this PR and doing the rest in a follow up?

tarunprabhu (approved):
The pointer_check tests have been a challenge to keep updated across platforms. I haven't had a chance to look at them closely, so thanks a lot for investigating.

I am ok with disabling the others as a follow-up if Tom doesn't mind.

tblah (approved):
Thanks for the follow-up

PR #484, opened 2026-09-11
[HIP] Migrate catch tests to Catch2 v3.8.1

kasaurov:
Upstream `hip-tests` requires `Catch2 v3`, and `v2` is no longer the branch where `Catch2` development happens.

Fetch the `v3.8.1` release tarball from `catchorg` rather than cloning. The previous `GIT_SHALLOW` clone pulled ~13MB, of which ~8MB was .git history discarded immediately, against ~1.2MB for the tarball. The download is pinned with `URL_HASH`.

`Catch2 v3` is a compiled library rather than a single header, so `CATCH2_INCLUDE_PATH` and the single_include path give way to linking `Catch2::Catch2`. That also removes a divergence between the `find_package` and `FetchContent` branches, which previously derived the include path in different ways.

Three details are not obvious from the diff:
* The generic-target helpers are built by a raw `clang++` command rather than a CMake target, so they take Catch2's include directories through a generator expression. The JOIN separator must be `$<SEMICOLON>`, since a literal ';' splits the string into list elements before the generator ever sees it.
* Those same commands need `-x none` before the `Catch2` archive, because the `-x hip` earlier on the line applies to every following input and clang would otherwise compile `libCatch2.a` as HIP source.
* The test-suite rewrites `CMAKE_CXX_COMPILE_OBJECT` to wrap compiles in `tools/timeit`, and that applies to Catch2's targets too. Nothing else orders them against `timeit`, so a clean build tree could race it.

No new tests are ported. Test output is unchanged; only the failure exit code differs, as `v3` returns a fixed `42` rather than v2's failed-assertion count. Nothing in the `lit verify` step or the summary script depends on the specific value.

### DIFF
--- a/External/HIP/HipCatchTests.cmake
+++ b/External/HIP/HipCatchTests.cmake
@@ -50,10 +50,18 @@ else()
   include(FetchContent)
   # Release tarball rather than a clone: the shallow clone pulls ~13MB, of which
   # ~8MB is .git history that is discarded immediately.
+  #
+  # The tarball and unpacked sources go to TEST_SUITE_HIP_MANAGED_DIR so they
+  # survive `rm -rf build/` instead of being re-downloaded. BINARY_DIR is left in
+  # the build tree on purpose: the compiled objects depend on the toolchain and
+  # build type, so sharing one across configurations would let them clobber
+  # each other.
   FetchContent_Declare(
     Catch2
-    URL      https://github.com/catchorg/Catch2/archive/refs/tags/v3.8.1.tar.gz
-    URL_HASH SHA256=18b3f70ac80fccc340d8c6ff0f339b2ae64944782f8d2fca2bd705cf47cadb79
+    URL          https://github.com/catchorg/Catch2/archive/refs/tags/v3.8.1.tar.gz
+    URL_HASH     SHA256=18b3f70ac80fccc340d8c6ff0f339b2ae64944782f8d2fca2bd705cf47cadb79
+    DOWNLOAD_DIR ${TEST_SUITE_HIP_MANAGED_DIR}/catch2/download
+    SOURCE_DIR   ${TEST_SUITE_HIP_MANAGED_DIR}/catch2/src
   )
   FetchContent_MakeAvailable(Catch2)
   # The test-suite rewrites CMAKE_CXX_COMPILE_OBJECT to wrap compiles in
@@ -286,10 +294,12 @@ function(create_generic_target_executables TEST_BASENAME TEST_DIR VARIANT_SUFFIX
       -unwindlib=libgcc
       -frtlib-add-rpath
       ${_include_flags}
-      # -x hip above applies to every following input, which would make clang
-      # try to compile the Catch2 archive as HIP source. Reset it first.
-      -x none
-      "$<TARGET_FILE:Catch2::Catch2>"
+      # Link Catch2 by -L/-l rather than by path: -x hip above applies to every
+      # following input file, and would make clang compile the archive as HIP
+      # source. The base name comes from the target so that it picks up
+      # DEBUG_POSTFIX, which makes the library libCatch2d in Debug builds.
+      -L$<TARGET_FILE_DIR:Catch2::Catch2>
+      -l$<TARGET_FILE_BASE_NAME:Catch2::Catch2>
       ${_libfs_flag}
     DEPENDS ${_common_sources}
     WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}"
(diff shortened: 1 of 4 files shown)


jplehr:
Is the shared storage that got introduced in https://github.com/llvm/llvm-test-suite/pull/421 a potential place where the download artifact can be stored?

External/HIP/HipCatchTests.cmake
```diff
@@ -308,11 +322,16 @@ function(create_generic_target_executables TEST_BASENAME TEST_DIR VARIANT_SUFFIX
       -unwindlib=libgcc
       -frtlib-add-rpath
       ${_include_flags}
+      # -x hip above applies to every following input, which would make clang
+      # try to compile the Catch2 archive as HIP source. Reset it first.
+      -x none
+      "$<TARGET_FILE:Catch2::Catch2>"
```
jmmartinez, line 328: If I understand, here we're passing the path to `libCatch2.a`, which is treated as a hip source thanks to the `-x hip`.

Would `-L $<TARGET_FILE_DIR:Catch2::Catch2> -lCatch2` work and avoid the `-x none`?

kasaurov:
> Is the shared storage that got introduced in #421 a potential place where the download artifact can be stored?

Good point, addressed by https://github.com/llvm/llvm-test-suite/pull/484/changes/c7d6628937db1d4963519880f4abaf17ed31fc8c

kasaurov:
> > Is the shared storage that got introduced in #421 a potential place where the download artifact can be stored?
> 
> Good point, addressed by [c7d6628](https://github.com/llvm/llvm-test-suite/commit/c7d6628937db1d4963519880f4abaf17ed31fc8c)

To be exact - `DOWNLOAD_DIR` and `SOURCE_DIR` now point at `${TEST_SUITE_HIP_MANAGED_DIR}/catch2`, so the tarball and unpacked sources survive `rm -rf build/`. 

Left `BINARY_DIR` at its default inside the build tree on purpose. Catch2's compiled objects depend on the toolchain and build type, so putting them in the shared location would let two configure trees clobber each other. Only the network-reconstructible parts go in the managed dir, which matches how that variable is documented.

External/HIP/HipCatchTests.cmake
kasaurov: Good idea! Adopted with one adjustment by https://github.com/llvm/llvm-test-suite/pull/484/changes/c7d6628937db1d4963519880f4abaf17ed31fc8c.
A literal `-lCatch2` would break Debug builds, because Catch2 sets `DEBUG_POSTFIX "d"` on the target, so the archive is `libCatch2d.a` there. Using the target-derived base name handles it:
```
-L$<TARGET_FILE_DIR:Catch2::Catch2>
-l$<TARGET_FILE_BASE_NAME:Catch2::Catch2>
```
Confirmed both cases: a Release configure generates `-lCatch2` and a Debug configure generates `-lCatch2d`. The `-x none` is gone, since `-L/-l` are not input files and `-x hip` doesn't apply to them.

jmmartinez: Thanks!

jplehr (approved):
lg

kovdan01:
Configuring llvm-test-suite with `External` selected fails after this PR on a builder without HIP prerequisites: see e.g. https://lab.llvm.org/buildbot/#/builders/227/builds/5017/steps/15/logs/stdio

`ENABLE_HIP_CATCH_TESTS` defaults to `OFF`, but `HipCatchTests.cmake` seems to perform the Catch2 lookup and fallback download unconditionally.

In the failing configuration, `TEST_SUITE_HIP_ROOT` is empty and `TEST_SUITE_HIP_MANAGED_DIR` is therefore set to `/managed`. Catch2 acquisition then fails trying to create `/managed/catch2/src`.

I suppose that Catch2 acquisition should probably be gated by `ENABLE_HIP_CATCH_TESTS`.

Could you please take a look at this and propose such a check so those not interested in HIP tests would not face this Catch2-related issue?

Also tagging @vvereschaka @asl @atrosinenko

DavidSpickett:
Many external CMakeLists.txt have a `if(TEST_SUITE_<name>_ROOT)` conditional before doing anything. Perhaps HIP can do the same.

jplehr:
@kasaurov can you look at the reported problem here and my (hopeful) functioning workaround in https://github.com/llvm/llvm-test-suite/pull/485

I'm sure what David suggested with the global conditional is a good direction to look into.

kasaurov:
> @kasaurov can you look at the reported problem here and my (hopeful) functioning workaround in #485
> 
> I'm sure what David suggested with the global conditional is a good direction to look into.

A follow up PR https://github.com/llvm/llvm-test-suite/pull/486

PR #485, opened 2026-09-15
[HIP] Properly guard HIP catch tests

jplehr:
This should fix a configuration issue after https://github.com/llvm/llvm-test-suite/pull/475 which led to a few places of HIP catch test configuration not being guarded properly.


mhalk (approved):
LGTM! Performed another sweep across the file and it seems like "catch"-occurrences are taken care of with these guards.

PR #486, opened 2026-09-16
[HIP] Return early when HIP prerequisites are missing

kasaurov:
External/HIP ran its entire file scope before checking `TEST_SUITE_HIP_ROOT`, which only guarded the final `create_hip_tests()` call. That is how `TEST_SUITE_HIP_MANAGED_DIR` came to be `"/managed"` on builders without HIP and broke configure after #484; #485 fixed the symptom by guarding the Catch2 acquisition.

Return early instead, as other suites do, so nothing can derive paths from an empty root. The trailing `if(TEST_SUITE_HIP_ROOT)`; is then redundant.

Updates the readme.


External/HIP/CMakeLists.txt
```diff
@@ -5,9 +5,22 @@ include(GPUTestVariant)
 llvm_externals_find(TEST_SUITE_HIP_ROOT "hip" "HIP prerequisites")
 message(STATUS "TEST_SUITE_HIP_ROOT: ${TEST_SUITE_HIP_ROOT}")
```
jplehr: Should this message now appear after we found that we want to run the HIP test suite, i.e., after the newly early-exit block?

jplehr (approved):
One nit

External/HIP/CMakeLists.txt
kasaurov: Good point! Addressed by https://github.com/llvm/llvm-test-suite/pull/486/changes/5d7827e12604491d3ba678e0665f02bd82ce4a22

jplehr:
@kasaurov should I land this?

kasaurov:
> @kasaurov should I land this?

Yes, please!

PR #487, opened 2026-09-16
[Fortran][gfortran] Clean up pointer_check_<N>.f90 test annotations

DavidSpickett:
Follow up to #483.

Flang does not implement -fcheck=pointer, and many of the tests include UB that's supposed to be caught by that feature.

1,2,3 and 4 "fail" by crashing, which could change in future to anything, so disable them, pointer_check_10 requires a standard we don't support.

A lof of them are "xfailed" in test.cmake, because update-test-config.py knows that Flang doesn't support the flag.

So we will have a disable in DisabledFiles.cmake, and an xfail in tests.cmake. Which is odd but I think that's the way it's supposed to work. The auto generated tests.cmake is overlaid by the other files, so the disables win.


tblah (approved):
LGTM, thanks

PR #488, opened 2026-09-16
[Fortran][gfortran] Ignore decode errors when reading test files

DavidSpickett:
Someone hit this on #91, and so did I. The problem was that achar_2.f90 has some characters that do not decode properly in the detected encoding.
```
>>> f = open('Fortran/gfortran/regression/achar_2.f90', 'rb')
>>> import chardet
>>> chardet.detect(f.read())
{'encoding': 'Windows-1254', 'confidence': 0.34330631676345713, 'language': 'Turkish'}
>>> f.close()
>>> f = open('Fortran/gfortran/regression/achar_2.f90', 'r', encoding='Windows-1254')
>>> f.readlines()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.10/encodings/cp1254.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 6415: character maps to <undefined>
```
This test is something to do with encoding, so it's not that surprising. There are no special encodings in the file headers that we need to parse, so I think it's safe to ignore decode errors. This returns the file content but with the invalid characters missing. See https://docs.python.org/3/builtins/functions.html#open

While I'm here, add the error text to the warning so problems are easier to find in future.


tblah (approved):
Thanks!

PR #489, opened 2026-09-16
[Fortran][gfortran] Disable enum_10.f90 because it causes UB in flang

DavidSpickett:
This has an enumerator that is the maximum value of it's type kind. This triggers UB in flang, which we detected using UBSAN.

Full details in https://github.com/llvm/llvm-project/issues/224030.

This hasn't caused problems in normal builds so far, but we are trying to bring up a santized flang
testing bot. So it would be ideal if we disable the test while until the problem is fixed.


DavidSpickett:
I know we have disabled tests because they themselves have UB in them, this is the first one that causes UB in flang instead.

tblah (approved):
Thanks

PR #490, opened 2026-09-17
[simd_ops] Update target layout for ARM bitcode files

DavidSpickett:
Since https://github.com/llvm/llvm-project/pull/224012, we do not overwrite the target layout of a module.
This lead to failures on our Arm bots:
```
fatal error: error in backend: Can't create a MachineFunction using a Module with a
      Target-incompatible DataLayout attached
  Target DataLayout: e-m:e-p:32:32-Fi8-i64:64-v128:64:128-a:0:32-n32-S64
  Module DataLayout: e-m:e-p:32:32-i64:64-v128:64:128-a:0:32-n32-S64
```
Clang's data layout has an extra "Fi8". Fi refers to the
alignment of function pointers, which is some multiple of 8.

We must have been compiling these bitcode files with clang's
target layout all along, so I think we're safe to update the
layout in the files to match clang's.

They all have the same layout:
```
$ cat scan.sh
for f in ARM_halide_runtime.bc ARM_tests/*.bc ARM_scalar_tests/*.bc; do
  ~/build-llvm-aarch64/bin/llvm-dis "$f" -o - 2>/dev/null |
    grep -m1 '^target datalayout'
done | sort | uniq -c
$ bash scan.sh
   2779 target datalayout = "e-m:e-p:32:32-i64:64-v128:64:128-a:0:32-n32-S64"
```
That 2779 is 2x1389 failing tests, plus the halide runtime file 
(each test is runtime plus 2 other files).

I disassembled them, changed the layout and re-assembled them:
```
for f in ARM_halide_runtime.bc ARM_tests/*.bc ARM_scalar_tests/*.bc; do
  tmp=$(mktemp --suffix=.ll)
  ~/build-llvm-aarch64/bin/llvm-dis "$f" -o "$tmp" || exit 1
  sed -i \ 
    's/e-m:e-p:32:32-i64:64-v128:64:128-a:0:32-n32-S64/e-m:e-p:32:32-Fi8-i64:64-v128:64:128-a:0:32-n32-S64/' \
    "$tmp"
  ~/build-llvm-aarch64/bin/llvm-as "$tmp" -o "$f" || exit 1
  rm "$tmp"
done
```
After doing this, the tests build and run correctly. The binary size 
changes more than I expected, but the disassembled content is the
same for both.
```
mkdir -p /tmp/simd-before
for f in ARM_halide_runtime.bc ARM_tests/*.bc ARM_scalar_tests/*.bc; do
  out="/tmp/simd-before/${f%.bc}.ll"
  mkdir -p "$(dirname "$out")"
  ~/build-llvm-aarch64/bin/llvm-dis "$f" -o "$out" || exit 1
done
```
Only target layout lines were added and removed:
```
$ diff -ru /tmp/simd-before /tmp/simd-after | grep "^-" | grep -v "^--" | sort | uniq 
-target datalayout = "e-m:e-p:32:32-i64:64-v128:64:128-a:0:32-n32-S64"
$ diff -ru /tmp/simd-before /tmp/simd-after | grep "^+" | grep -v "^++" | sort | uniq 
+target datalayout = "e-m:e-p:32:32-Fi8-i64:64-v128:64:128-a:0:32-n32-S64"
```


DavidSpickett:
I think this is preferable to regenerating the files, but I can do that if it's the right thing to do.

There is https://github.com/llvm/llvm-test-suite/blob/main/Bitcode/README_HalideCodegen but it refers to top of trunk Halide from 10 years ago, and I think I would have to update the other targets' files too.

The size of the bitcode files has increased a lot, I could try getting a 10 year old set of tools (~llvm 4.0) to do the rewrite with instead.

Github might struggle to render anything, so here's a sample of the change:
```
 Bitcode/simd_ops/ARM_halide_runtime.bc                                | Bin 108644 -> 130816 bytes
 Bitcode/simd_ops/ARM_scalar_tests/scalar_test_op_vaba_s16_3.bc        | Bin 33008 -> 43848 bytes
 Bitcode/simd_ops/ARM_scalar_tests/scalar_test_op_vaba_s16_331.bc      | Bin 33104 -> 43944 bytes
 Bitcode/simd_ops/ARM_scalar_tests/scalar_test_op_vaba_s16_662.bc      | Bin 33104 -> 43944 bytes
 Bitcode/simd_ops/ARM_scalar_tests/scalar_test_op_vaba_s16_991.bc      | Bin 33104 -> 43944 bytes
 Bitcode/simd_ops/ARM_scalar_tests/scalar_test_op_vaba_s32_333.bc      | Bin 33132 -> 43972 bytes
 Bitcode/simd_ops/ARM_scalar_tests/scalar_test_op_vaba_s32_5.bc        | Bin 33036 -> 43876 bytes
 Bitcode/simd_ops/ARM_scalar_tests/scalar_test_op_vaba_s32_664.bc      | Bin 33132 -> 43972 bytes
 Bitcode/simd_ops/ARM_scalar_tests/scalar_test_op_vaba_s32_993.bc      | Bin 33132 -> 43972 bytes
 Bitcode/simd_ops/ARM_scalar_tests/scalar_test_op_vaba_s8_1.bc         | Bin 32912 -> 43752 bytes
 Bitcode/simd_ops/ARM_scalar_tests/scalar_test_op_vaba_s8_329.bc       | Bin 33004 -> 43840 bytes
 Bitcode/simd_ops/ARM_scalar_tests/scalar_test_op_vaba_s8_660.bc       | Bin 33004 -> 43840 bytes
 Bitcode/simd_ops/ARM_scalar_tests/scalar_test_op_vaba_s8_989.bc       | Bin 33004 -> 43840 bytes
 Bitcode/simd_ops/ARM_scalar_tests/scalar_test_op_vaba_u16_332.bc      | Bin 33104 -> 43944 bytes
```
For example `scalar_test_op_vaba_s16_3.bc` is 32% larger.

arsenm:
I find the size increase quite surprising

DavidSpickett:
Actually in the files there is:
```
clang version 3.9.0 (trunk 275577) (llvm/trunk 275433)
```
I will try to build that.

arsenm:
Alternatively could handle this case in Bitcode auto upgrade

DavidSpickett:
The old tools change the size much less:
```
$ ~/build-old-llvm/bin/llvm-dis ARM_halide_runtime.bc -o /tmp/ARM_halide_runtime.ll
$ ~/build-old-llvm/bin/llvm-as /tmp/ARM_halide_runtime.ll -o /tmp/ARM_halide_runtime.bc
$ ls -lah ARM_halide_runtime.bc 
-rw-rw-r-- 1 davspi01 davspi01 107K Sep 17 14:58 ARM_halide_runtime.bc
$ ls -lah /tmp/ARM_halide_runtime.bc 
-rw-rw-r-- 1 davspi01 davspi01 109K Sep 17 15:04 /tmp/ARM_halide_runtime.bc
```
About 1.8%. However they do not accept `Fi8`:
```
$ ~/build-old-llvm/bin/llvm-dis ARM_halide_runtime.bc -o /tmp/ARM_halide_runtime.ll
$ vim /tmp/ARM_halide_runtime.ll 
$ head /tmp/ARM_halide_runtime.ll 
; ModuleID = 'ARM_halide_runtime.bc'
source_filename = "/usr/local/google/home/asbirlea/Halide/Halide/src/runtime/posix_allocator.cpp"
target datalayout = "e-m:e-p:32:32-Fi8-i64:64-v128:64:128-a:0:32-n32-S64"
<...>
$ ~/build-old-llvm/bin/llvm-as /tmp/ARM_halide_runtime.ll -o /tmp/ARM_halide_runtime.bc
LLVM ERROR: Unknown specifier in datalayout string
```

> Alternatively could handle this case in Bitcode auto upgrade

Yes, 3.9 is >3.0 so I think https://www.llvm.org/docs/DeveloperPolicy.html?utm_source=chatgpt.com#ir-backwards-compatibility applies. I see that we already have `UpgradeDataLayoutString`.

I'll do that instead. Thanks for the pointer.

DavidSpickett:
> I'll do that instead.

https://github.com/llvm/llvm-project/pull/224639

PR #492, opened 2026-09-21
[Vectorizer] Test runtime checks for stencil access patterns

igogo-x86:
Compare scalar and vector execution for overlapping stencil accesses with one or two runtime strides and uint8_t, uint32_t, and uint64_t elements. Add a forced stencil-merging variant when the compiler supports it.

Assisted by: ChatGPT


SingleSource/UnitTests/Vectorizer/runtime-checks.cpp
```diff
@@ -153,6 +153,92 @@ static void checkOverlappingMemoryTwoRuntimeChecksNested(Fn4Ty<Ty> ScalarFn,
...
+                                            "uint64_t", S, 0);
+  }
+}
+
+// Check all element types with two strides. Try each pair in both orders,
+// because either stride can be the larger one.
+template <typename ScalarFnTy, typename VectorFnTy>
+static void checkTwoStrides(ScalarFnTy ScalarFn, VectorFnTy VectorFn,
+                            const char *Name) {
+  std::pair<long, long> Strides[] = {{1, 2},    {2, 1},    {7, 16}, {16, 7},
```
david-arm, line 231: Do we need to include the relevant C++ header for `std::pair`?

SingleSource/UnitTests/Vectorizer/runtime-checks.cpp
```diff
@@ -153,6 +153,92 @@ static void checkOverlappingMemoryTwoRuntimeChecksNested(Fn4Ty<Ty> ScalarFn,
     CheckWithOffsetSecond(i);
 }
 
+#define DEFINE_SCALAR_AND_VECTOR_STENCIL_FN(Loop)                              \
+  auto ScalarFn = [](auto *Out, auto *In, unsigned TC, long S1, long S2) {     \
```
david-arm, line 157: Is it worth testing different stride types, given that https://github.com/llvm/llvm-project/pull/187252 may behave differently for uint8_t strides, etc?

SingleSource/UnitTests/Vectorizer/runtime-checks.cpp
david-arm, line 231: What about a negative test for negative strides?

SingleSource/UnitTests/Vectorizer/runtime-checks.cpp
```diff
@@ -153,6 +153,92 @@ static void checkOverlappingMemoryTwoRuntimeChecksNested(Fn4Ty<Ty> ScalarFn,
...
+  auto VectorFn = [](auto *Out, auto *In, unsigned TC, long S1, long S2) {     \
+    _Pragma("clang loop vectorize(enable)") Loop                               \
+  };
+
+template <typename Ty>
+using StencilFnTy = std::function<void(Ty *, Ty *, unsigned, long, long)>;
+
+template <typename Ty>
+static void
+checkOverlappingMemoryStencil(StencilFnTy<Ty> ScalarFn,
```
david-arm, line 169: Is there at least one negative test where your new stencil algorithm bails out? It would be good to have some here I think.

SingleSource/UnitTests/Vectorizer/runtime-checks.cpp
igogo-x86: Yeah, this file has already missed several `inlcude`s.
I started conservatively by not adding anything, but since I am adding new, I will as well fix that by adding non-present before.

SingleSource/UnitTests/Vectorizer/runtime-checks.cpp
igogo-x86: Yep, added uint8_t, uint32_t and uint64_t and uint8_t tests overflow in some of them.

SingleSource/UnitTests/Vectorizer/runtime-checks.cpp
igogo-x86: Added negative strides -16, -1 and also 0 (and -128 for 8-byte stride)

SingleSource/UnitTests/Vectorizer/runtime-checks.cpp
igogo-x86: Added predicated access test `i % 3 != 0`

david-arm (approved):
LGTM!

